PR #11, opened 2022-06-12
[#3] Augment websockets loop

### LINKED ISSUE #3: Websockets main loop
**UPDATE**: 

It seems there's a (work in progress) library that implements an interface for Slack's SocketMode called [`slacker`](https://github.com/velveteer/slacker). It implements some of the features we need (e.g. handle `"disconnect"` events).

We should try replacing our websocket loop with `slacker`.

----

Right now, we have a websockets application that consists of a simple loop (see `TzBot.Slack.SocketMode.Client.wsApp`): it waits for a message, reads it as `Text`, prints it to stdout, repeat.

Here's what the loop should do:

- [ ] Receive and parse a Slack Event
- [ ] Send an "acknowledgement message" for each received message (see [docs](https://api.slack.com/apis/connections/socket-implement#acknowledge))
- [ ] If we receive an event with `"type": "message"` (see the next section), we construct a `EventSummary` and run `processEvent` asynchronously.
- [ ] If we receive an event with`"type": "disconnect"` (see [docs](https://api.slack.com/apis/connections/socket-implement#disconnect)), exit the loop, establish a fresh websockets connection and re-start the loop.
- [ ] All other events are ignored.

--- 

### Parsing events

Example jsons:
 * A new message was posted to a channel: https://gist.github.com/dcastro/b560d6fcae173de64cd97f17587faab6
 * A message has been edited: https://gist.github.com/dcastro/6e4d9fa3e606aeb0f74ec7cd1531558e

All messages delivered via websockets (aka Slack's "Socket Mode") are wrapped in an "envelope" ([docs](https://api.slack.com/apis/connections/socket-implement#events)).

Inside that envelope, you'll find an "event wrapper", whose format is described [here](https://api.slack.com/types/event).

Inside that event wrapper, we'll find the actual event info. In this case, we're only interested in "message" events (see [docs](https://api.slack.com/events/message)).

Some of those "message" events may have an optional `"subtype": "message_changed"`, indicating that a message has been edited (see [docs](https://api.slack.com/events/message/message_changed)). Those events will have a `previous_message` field with the "old text" before it was edited. The docs seem to be incomplete in this regard, they don't mention this field, but you can see it in the example jsons above.

Note that we don't have to parse _all_ the fields. Only the ones necessary to build an `EventSummary`.

# Acceptance criteria

* We have a websockets loop capable of dispatching to `processEvent` when a new event is received.
* It handles requests to disconnect gracefully.

ixahedron:
Problem: our websockets app doesn't currently handle the incoming Slack events in any useful way. Additionally, we are trying to implement the AI ourselves, whereas there is an external library already doing that.

Solution: changed the websockets loop to use the 'slacker' lib, added extracting useful data from any message events, as well as some safeguards.

## Related issue(s)

Fixed #3

- Tests
        silently reappearing again.

- Documentation
    - Haddock


dcastro:
I'm getting this error when compiling with stack, I'm guessing due to the changes made in `stack.yaml`:

```hs
$ stack build
error: attribute 'ghc902' missing

       at «string»:1:48:

            1| with (import <nixpkgs> {}); let inputs = [zlib haskell.compiler.ghc902 git gcc gmp]; libPath = lib.makeLibraryPath inputs; stackExtraArgs = lib.concatMap (pkg: [ ''--extra-lib-dirs=${lib.getLib pkg}/lib''   ''--extra-include-dirs=${lib.getDev pkg}/include'' ]) inputs; in runCommand ''myEnv'' { buildInputs = lib.optional stdenv.isLinux glibcLocales ++ inputs; STACK_PLATFORM_VARIANT=''nix''; STACK_IN_NIX_SHELL=1; LD_LIBRARY_PATH = libPath;STACK_IN_NIX_EXTRA_ARGS = stackExtraArgs; LANG="en_US.UTF-8";} ""
             |                                                ^
(use '--show-trace' to show detailed location information)
```

Is there anything I need to do?

pasqu4le:
Ah, I have the same problem, it looks like the issue is with the `enable: true` part under the `nix` key that was added here.

Not sure if removing that disrupts @ixahedron's workflow, but I'd imagine that it won't.

src/TzBot/ProcessEvent.hs
```diff
@@ -20,7 +24,31 @@ data EventSummary = EventSummary
...
+constructEventSummary :: Data.Aeson.Value -> EventSummary
+constructEventSummary evt = EventSummary message previousMessage uid (extractChannelId evt) (extractTimeStamp evt)
+  where previousMessageObj = evt ^? key "previous_message" . _Value
+        previousMessage = evt ^? key "previous_message" . key "text" . _String
+
+        message = fromJust $ case previousMessage of
+          Nothing -> evt ^? key "text" . _String
+          Just _  -> evt ^? key "message" . key "text" . _String
+
+        -- We always extract the user id of the original message in case of forced edits
```
dcastro, line 48: Good catch!

PR #12, opened 2022-06-13
[#1] Add time references parser

### LINKED ISSUE #1: Parse time references

We need to be able to parse references to points in time from Slack messages.

See the [`parseTimeRefs`](https://github.com/serokell/tzbot/blob/main/src/TzBot/Parser.hs) and [`TimeReference`](https://github.com/serokell/tzbot/blob/755c2c62339a9d9ced89cfd20a29de34b5564d09/src/TzBot/TimeReference.hs#L29-L51)

A `TimeReference` should always contain a "time of day", which can be expressed using a 12-hour format:

> Does **10am** work for you?

> That doesn't work for me, what about **10:30 AM**?

Or a 24-hour format:

> I can only be there at 16:00

----

It _may_ also contain a reference to a certain day.

That reference can be expressed as:
* "today" / "tomorrow" / "yesterday" 
  > 10am tomorrow
  > today at 3pm
* a number of days from today
  > 10am in 2 days
* a day of the week
  > tuesday at 3pm
  > at 3pm on tuesday
* a day of the month
  > at 11am on the 4th 
* a day of the month + a month
  > at 11am on the 4th of April
  > at 11am on April 4
  > at 11am on 4 April

-----

It _may_ also contain a reference to a certain timezone **or** timezone abbreviation **or** offset.

* Timezone. See [`tzNameLabelMap`](https://hackage.haskell.org/package/tz-0.1.3.6/docs/Data-Time-Zones-All.html#v:tzNameLabelMap) for a list of all available timezone names.
  > 9am in europe/london
  > 
* Timezone abbreviation. For simplicity, we should consider any all-uppercase 2-5 letter word a timezone abbreviation (except `AM` and `PM`, of course). Validation of the timezone abbreviation will happen in a separate issue.
  > 2pm CST
* Offset.
  > 10am UTC+03:00
  > 10am UTC+3

----- 

A slack message may contain more than one reference, e.g.:

> Let's meet between 10am and 11:30am

# Acceptance criteria

* The `parseTimeRefs` function has been implemented and is able to parse the examples above.
* We don't need to be able to support all examples in one go - it's fine to submit a PR with support for _some_ cases, and then iterate on that.

pasqu4le:
## Description

This implement `parseTimeRefs` to support a wide (but possibly not complete) array of time references from the given Text.

All the examples in the issues have been added as doctests.

## Related issue(s)

Fixed #1

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/TzBot/Parser.hs
+++ b/src/TzBot/Parser.hs
@@ -4,8 +4,25 @@
 
 module TzBot.Parser where
 
+
+import Data.Char (digitToInt)
+import Data.Map (toList)
+import Data.Maybe (fromMaybe, isJust)
 import Data.Text (Text)
-import TzBot.TimeReference (TimeReference)
+import Data.Text.Encoding (decodeUtf8)
+import Data.Time (DayOfWeek(..))
+import Data.Time.Calendar.Compat (DayOfMonth, MonthOfYear)
+import Data.Time.LocalTime (TimeOfDay(..))
+import Data.Time.Zones.All (tzNameLabelMap)
+import Data.Void (Void)
+import Text.Megaparsec
+  (Parsec, anySingle, choice, empty, match, optional, parseMaybe, takeRest, try, (<|>))
+import Text.Megaparsec.Char (char, digitChar, space, space1, string')
+import Text.Megaparsec.Char.Lexer (decimal)
+import TzBot.TimeReference
+
+
+type TzParser = Parsec Void Text
 
 {- | Parses time references from an input string.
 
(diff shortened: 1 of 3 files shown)


src/TzBot/Parser.hs
```diff
@@ -4,24 +4,287 @@
...
+    ]
+
+dayOfMonthParser :: TzParser DayOfMonth
+dayOfMonthParser = do
+  n <- decimal
+  optional' (string' "st" <|> string' "nd" <|> string' "rd" <|> string' "th")
+  if n < 32 then pure n else empty
+
+monthOfYearParser :: TzParser MonthOfYear
+monthOfYearParser = choice' $
```
ixahedron, line 214: Right now it will accept dates like "31st of February", which isn't ever correct and should probably be reported to the user. I don't think it's important at this stage/for this parser, but we should think about handling it correctly.

ixahedron (approved):
LGTM!

PR #20, opened 2022-12-20
[#2] Convert time references to UTC

### LINKED ISSUE #2: Convert time references to UTC

Implement the [`timeReferenceToUTC`](https://github.com/serokell/tzbot/blob/9a39e77119d7ec02377075ca5100eb6a0000bf80/src/TzBot/TimeReference.hs#L62-L85) function.

Given a [`TimeReference`](https://github.com/serokell/tzbot/blob/9a39e77119d7ec02377075ca5100eb6a0000bf80/src/TzBot/TimeReference.hs#L29-L50) (see #1 for examples), we want to convert it to a moment in time expressed in UTC.

For example, if:
* a slack message is sent at 2022-05-01 11:00 UTC
* the sender is in the Asia/Tokyo timezone, which at that time is observing the offset UTC+09:00
* the message's text contains "can we meet at 8am tomorrow?", which we parse as:
    ```hs
    TimeReference (TimeOfDay 8 0 0) (Just (DaysFromToday 1)) Nothing
    ```

The result of `timeReferenceToUTC` should be 2022-05-01 23:00 UTC.

```hs
import Data.Time.Format.ISO8601
import Data.Time.LocalTime
import qualified Data.Time.Zones as TZ
import qualified Data.Time.Zones.All as TZ

λ> iso8601ParseM @_ @LocalTime "2022-05-02T08:00:00" <&> TZ.localTimeToUTCFull (TZ.tzByLabel TZ.Asia__Tokyo)
LTUUnique {_ltuResult = 2022-05-01 23:00:00 UTC, _ltuZone = JST}
```

-----

When there is no `LocationReference` (as in the example above), we should assume the time reference is relative to the sender's timezone.

When there is no `DateReference`, `TimeReference (TimeOfDay 10 0 0) Nothing Nothing` should be interpreted as "10:00 in the current day _in the given timezone_".
If it's already past 10:00 in the given timezone, it should be interpreted as "10:00 _the next day_ in the given timezone".

When `DateReference` is `DaysFromToday n`, we should interpret that as `n` days from the current date _in the given timezone_.

When `DateReference` is a day of the week, e.g. `Tuesday`, it should be interpreted as _the next tuesday_ (or today, if today is tuesday _in the given timezone_).

-------

Be aware that:

1. Some date/times are ambiguous (see [Edge cases & pitfalls](https://github.com/serokell/tzbot/blob/main/docs/pitfalls.md) for more details). When this happens, `timeReferenceToUTC` should return `TRTUAmbiguous`
2. Some date/times are invalid. When this happens, it should return `TRTUInvalid`
3. If the sender uses a timezone abbreviation that we don't support / doesn't exist (e.g. "10am in XYZ"), it should return `TRTUInvalidTimeZoneAbbrev`.

# Acceptance criteria

* `timeReferenceToUTC` is capable of transforming time references into UTC datetimes.
* It's capable of handling and detecting ambiguous and invalid time references.
* It's capable of handling invalid timezone abbreviations.

YuriRomanowski:
Problem: currently we can obtain time references via parsing but are not able to convert this references to a timestamp.

Solution: implement `timeReferenceToUTC` function
that converts time reference to a timestamp, possibly resulting in an error.

## Related issue(s)

Fixed #2

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/TzBot/TimeReference.hs
+++ b/src/TzBot/TimeReference.hs
@@ -6,10 +6,14 @@ module TzBot.TimeReference where
 
 import Universum
 
-import Data.Maybe (fromJust)
-import Data.Time (DayOfWeek, TimeOfDay, TimeZone, UTCTime)
+import Control.Arrow ((>>>))
+import Data.List.NonEmpty qualified as NE
+import Data.Time
+  (DayOfWeek, LocalTime(localDay), NominalDiffTime, TimeOfDay, TimeZone, UTCTime, addLocalTime,
+  diffLocalTime, localTimeToUTC, nominalDay, toGregorian, utc, utcToLocalTime)
 import Data.Time.Calendar.Compat (DayOfMonth, MonthOfYear)
-import Data.Time.Format.ISO8601 (iso8601ParseM)
+import Data.Time.TZInfo qualified as TZI
+import Data.Time.TZTime qualified as TZT
 import Data.Time.Zones.All (TZLabel)
 import Formatting (Buildable)
 
@@ -18,7 +22,7 @@ import Formatting (Buildable)
 Note: The `TimeZone` data type from the @time@ package is a misnomer, it doesn't actually represent a timezone.
 
 A timezone contains a set of rules dictating which offset(s) is/are observed throughout the year.
-For example: under current law, the “Europe/London” timezone observes the offset BST (UTC+01:00)
+For example: under current law, the "Europe/London" timezone observes the offset BST (UTC+01:00)
 during summer and the offset GMT (UTC+00:00) otherwise.
 These rules change over time by governmental decree.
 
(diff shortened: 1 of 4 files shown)


src/TzBot/TimeReference.hs
```diff
@@ -62,20 +67,103 @@ newtype TimeZoneAbbreviation = TimeZoneAbbreviation { unTimeZoneAbbreviation ::
...
+        else TZT.addCalendarClip (TZT.calendarDays 1) eventLocalTime
+    Just (DaysFromToday n) ->
+      TZT.addCalendarClip (TZT.calendarDays $ fromIntegral @Int @Integer n) eventLocalTime
+    Just (DayOfWeekRef dayOfWeek) ->
+      TZT.atFirstDayOfWeekOnAfter dayOfWeek eventLocalTime
+    Just (DayOfMonthRef dayOfMonth mbMonthOfYear) -> case mbMonthOfYear of
+        -- Currently for both cases we don't handle invalid days of
+        -- month (just clip them to valid ones)
+        -- and don't consider days in the past
+        -- so if now is 30 Sep and the user mentioned "31th" (without month), we should get 30 Sep
```
YuriRomanowski, line 142: This code means, that if today is, say, 2 September 2022, and the time reference contains 1 September, our bot will infer 1 September 2023. This behavior is rather strange, but I can hardly imagine that I mention date in the past with exact time, so maybe that's not a big problem. We could implement more complex logic, for example pick the 1 September that is closer to the current moment, even if it's in the past, if it's required. What do you think?

src/TzBot/TimeReference.hs
```diff
@@ -129,8 +217,8 @@ knownTimeZoneAbbreviations =
   , TimeZoneAbbreviationInfo "AMST"  (hours -03)       "Amazon Summer Time"
   , TimeZoneAbbreviationInfo "CLT"   (hours -04)       "Chile Standard Time"
   , TimeZoneAbbreviationInfo "CLST"  (hours -03)       "Chile Summer Time"
-  , TimeZoneAbbreviationInfo "BRT"   (hours -03)       "Brasília Time"
-  , TimeZoneAbbreviationInfo "BRST"  (hours -02)       "Brasília Summer Time"
+  , TimeZoneAbbreviationInfo "BRT"   (hours -03)       "Brasilia Time"
+  , TimeZoneAbbreviationInfo "BRST"  (hours -02)       "Brasilia Summer Time"
```
YuriRomanowski, line 221: `stylish` wasn't able to devour these special characters painlessly.

src/TzBot/TimeReference.hs
Martoon-00: Oh, hard things need discussion here.

My personal opinion is that a timestamp referring to yesterday or maybe a week ago is possible ("Could you remind when you had a meeting with Z?"), but further in the past - unlikely, as you said it.

I also think that the user hardly ever means a day that is 354 days ahead, timestamps for days that are many months ahead do not make sense even if it is some planned ZuriHac event - time is usually clarified closer to the target date because anything can happen.

And I actually think about one trick to use here and in many other places: e.g. for any date outside of 
`today + [- 1 week; +3 months]` range we could say the sender: sorry, your date is ambiguous, please add the year explicitly; and that would be valid, because other people may also find the same message ambiguous.

What do you think?

Martoon-00: Also one suggestion: there is generally a practice to extract magic constants to separate variables. This way one needs not to read the entire code to tune up the constants, constants are easier to extract to configuration, e.t.c.

Maybe whenever we see such points of ambiguity (e.g. how to infer the year), we could extract _entire rules (i.e. functions)_ into separate variables. This way the main logic will be rigid, and all the questionable moments will appear separately and will be easy to update if necessary.

No need to strictly follow this suggestion, but maybe you'll find it sensible too and in some situations it will be easy to apply.

src/TzBot/TimeReference.hs
```diff
@@ -4,20 +4,25 @@
...
+import Data.Time.TZTime qualified as TZT
+import Data.Time.Zones.All (TZLabel(..))
 
 {- | An offset from UTC (e.g. @UTC+01:00@) with an optional timezone abbreviation (e.g. @BST@).
 
 Note: The `TimeZone` data type from the @time@ package is a misnomer, it doesn't actually represent a timezone.
 
 A timezone contains a set of rules dictating which offset(s) is/are observed throughout the year.
-For example: under current law, the “Europe/London” timezone observes the offset BST (UTC+01:00)
+For example: under current law, the "Europe/London" timezone observes the offset BST (UTC+01:00)
```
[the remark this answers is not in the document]
YuriRomanowski: Yes, my `stylish` just removed everything after the first quote...

src/TzBot/TimeReference.hs
YuriRomanowski: > Maybe whenever we see such points of ambiguity (e.g. how to infer the year), we could extract entire rules (i.e. functions) into separate variables. This way the main logic will be rigid, and all the questionable moments will appear separately and will be easy to update if necessary.

I just factored out functions `chooseBestMonth` and `chooseBestYear`, did you mean that?

YuriRomanowski: > And I actually think about one trick to use here and in many other places: e.g. for any date outside of
today + [- 1 week; +3 months] range we could say the sender: sorry, your date is ambiguous, please add the year explicitly; and that would be valid, because other people may also find the same message ambiguous.

After the new commit, we choose the best suiting option: for only day of month known, we try this months and adjacent ones, picking the closest, but preferring future days over past days with some handicap (named as `preferFuture` in code). Also we try to find exactly matching date with minimal weight, if there is none, we take just the day with minimal weight. What do you think?

About your suggestion, we can add one more field to `TRTUSuccess`, indicating that we are not sure about the date we inferred: either it's too far from today (outside your interval, I think it's OK), or the day of month is not matching. But it's not really an error, so we can just indicate that this date is possibly ambiguous in ephemeral messages.

test/Test/TzBot/TimeReferenceToUtcSpec.hs
```diff
@@ -0,0 +1,165 @@
...
+time3 :: UTCTime
+time3 = toUTC [tz|2022-11-04 10:30:00 [America/Havana]|]
+
+labelHavana :: TZLabel
+labelHavana = America__Havana
+
+hour :: Int
+hour = 60
+
+mkTestCase :: (TimeReference, TZLabel, UTCTime, TimeReferenceToUTCResult) -> Assertion
```
alexd1971, line 36: Why do you use tuple instead of normal function parameters?

alexd1971 (approved):
Approved with a question and suggestion

test/Test/TzBot/TimeReferenceToUtcSpec.hs
YuriRomanowski: Hm, the original idea was to define a list of tuples and then `map` them, but I think you are right, here this idea is not used.
Maybe even better it would be to define a datastructure such that every input parameter had a name when calling `mkTestCase`

src/TzBot/TimeReference.hs
Martoon-00: > I just factored out functions chooseBestMonth and chooseBestYear, did you mean that?

Yep. The result looks awesome :+1: 

> we can add one more field to TRTUSuccess, indicating that we are not sure about the date we inferred

Hmm, that's also an option to consider. Let's discuss this elsewhere I guess.

Martoon-00: Created #36.

PR #21, opened 2022-12-22
[#4] Process slack events

### LINKED ISSUE #4: Process Slack events

We should implement `TzBot.ProcessEvent.processEvent`.

This function coordinates the work done on other issues (and can be done independently).

It receives an `EventSummary` and then:
- [ ] Calls `parseTimeRefs` (implemented in #1) to extract time references from a message
- [ ] Use `timeReferenceToUTC` (implemented in #2) to convert them to UTC datetimes.
    * This returns two things:  the time reference converted to UTC (let's call it `timeRefUTC`) and the timezone OR the offset it's related to (let's call it `timeRefLocation`).
- [ ] Use `getChannelMembers` to get a list of all channel members.
- [ ] Process each user asynchronously (hint: use the `lifted-async` package), and send an ephemeral message to the user using `postEphemeral`.

## "Message edited" events

When the `EventSummary` indicates a Slack message has been edited, we should compare the time references in the "old text" against the time references in the "new text", and process only new ones.

For example, if a slack message has been edited from:

> Let's meet between 10am and 11:30am

To:

> Let's meet between 10am and 12:30am

We'll only want to process the "12:30am" reference.

## Rendering an ephemeral message

The message we send to the user should look something along these lines:

> [TimeRef.trText] is [x] in [user's timezone]

`[x]` can be obtained by converting `timeRefUTC` to the user's timezone using [`utcToLocalTimeTZ`](https://hackage.haskell.org/package/tz-0.1.3.6/docs/Data-Time-Zones.html#v:utcToLocalTimeTZ)

If the `TimeReference.trLocationRef` is empty (i.e. if it does not explicitly mention a timezone/offset/timezone abbreviation), we should add the sender's timezone to the message:

> [TimeRef.trText] in [sender's timezone] is [x] in [user's timezone]

## Skipping users

We don't always want to send an ephemeral message to _all_ users. If a user meets one of the following criteria, we should skip it:

* if the user is a bot
* If `timeRefLocation` is a timezone, and if the user is in that same timezone.
* ~If `timeRefLocation` is an offset, and if, at the time of `timeRefUTC`, the user's timezone will be observing that same offset.~
   * ~Note: use [`timeZoneForUTCTime`](https://hackage.haskell.org/package/tz-0.1.3.6/docs/Data-Time-Zones.html#v:timeZoneForUTCTime) to find out which offset the user will be observing at that time in their timezone.~

YuriRomanowski:
## Description

Problem: `processEvents` function should be implemented to make bot work.

Solution: implement `processEvents`. Handle ordinary messages, messages
in threads, edited messages (check only new time references).

## Related issue(s)

Fixed #4 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


package.yaml
```diff
@@ -19,25 +19,30 @@ description:         Please see the README on GitHub at <https://github.com/sero
...
   - megaparsec
+  - monad-control
   - mtl
   - nyan-interpolation
-  - safe-exceptions
   - servant-auth
   - servant-auth-client
   - servant-client
   - servant-server
+  - slacker
```
Martoon-00, line 39: Oh, that's interesting

src/TzBot/RunMonad.hs
```diff
@@ -18,6 +30,7 @@ data BotConfig = BotConfig
 data BotState = BotState
   { wasConfig :: BotConfig
   , wasManager :: Manager
+  , wasMessagesReferences :: IORef (M.Map MessageId (S.Set TimeReferenceText))
```
alexd1971, line 33: What does prefix `was` mean?

src/TzBot/BotMain.hs
```diff
@@ -38,17 +37,18 @@ main = do
             & setAppToken appLevelToken
             & setOnException handleThreadExceptionSensibly -- auto-handle disconnects
 
-  let wCfg = WebAPIConfig {
+  let wCfg = BotConfig {
     wacAppLevelToken = AppLevelToken appLevelToken
   , wacBotToken = BotToken botToken
```
alexd1971, line 44: I guess prefixies also should be renamed

src/TzBot/RunMonad.hs
```diff
@@ -0,0 +1,61 @@
...
+newtype BotToken = BotToken { unBotToken :: Text }
+
+data BotConfig = BotConfig
+  { wacAppLevelToken :: AppLevelToken
+  , wacBotToken :: BotToken
+  }
+
+data BotState = BotState
+  { wasConfig :: BotConfig
+  , wasManager :: Manager
```
alexd1971, line 32: these prefixes also seem to be obsolete

src/TzBot/RunMonad.hs
```diff
@@ -0,0 +1,61 @@
...
+newtype BotM a = BotM
+  { unBotM :: ReaderT BotState (ExceptT BotException IO) a
+  }
+  deriving newtype
+    ( Functor, Applicative, Monad
+    , MonadReader BotState, MonadError BotException
+    )
+
+runBotM :: BotState -> BotM a -> IO (Either BotException a)
+runBotM state webApiM =
```
alexd1971, line 46: I propose for consistency rename everywhere `webApiM` to `botM`

src/TzBot/ProcessEvent.hs
```diff
@@ -2,55 +2,83 @@
...
+filterNewReferencesAndMemorizePure :: [TimeReference] -> S.Set TimeReferenceText -> (S.Set TimeReferenceText, [TimeReference])
+filterNewReferencesAndMemorizePure newRefs refsSet = foldl' f (refsSet, []) newRefs
+  where
+    f (accSet, accRefs) ref = do
+      let newAccSet = S.insert (trText ref) accSet
+      if S.size newAccSet == S.size accSet
+        then (newAccSet, accRefs)
+        else (newAccSet, ref : accRefs)
+
+-- | When an edited message comes, define what time references in it haven't appear before
```
alexd1971, line 34: Why this function has haddock comment format but functions before and after don't?

alexd1971 (approved):
I left two comments but this doesn't prevent me to approve.

src/TzBot/ProcessEvent.hs
YuriRomanowski: I think that `filterNewReferencesAndMemorizePure` is just a helper function for `filterNewReferencesAndMemorize`, it could be defined as internal function for the latter one, but I liked it more as a separate function. In fact there's nothing more to say about this helper after `filterNewReferencesAndMemorize`.

src/TzBot/ProcessEvent.hs
```diff
@@ -2,55 +2,83 @@
...
+-- We don't need to handle MDMessageBroadcast anyhow,
+-- because we were supposed to reply to the original message
+-- in the thread earlier.
+newMessageOrEditedMessage :: MessageEvent -> Bool
+newMessageOrEditedMessage evt = case meMessageDetails evt of
+  MDMessage          -> True
+  MDMessageEdited {} -> True
+  MDMessageBroadcast -> False
+
+processEvent :: MessageEvent -> BotM ()
```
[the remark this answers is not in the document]
YuriRomanowski: I thought that this function is going to change a lot of times in the near future :) But let's add some temporary docs right now.

src/TzBot/ProcessEvent.hs
alexd1971: I don't actually know does it make sense to make haddock comment to functions which are not exported from the module?

YuriRomanowski: > I don't actually know does it make sense to make haddock comment to functions which are not exported from the module?

It's rather convenient when using HLS, because you can see the docs without going to the function itself. From other points of view it's useless.

alexd1971: But I guess if the function is not exported from a module you can't use it anywhere else. And it you don't need to look at function description. But if you are in the module you probably don't need haddock. Or I messed something up?

YuriRomanowski: Of course, but this can make sense for big modules. But yes, it seems it's completely redundant. If you like, I can delete '|' symbol, converting haddocks into ordinary comments.

PR #23, opened 2022-12-27
[#6] Add configuration

### LINKED ISSUE #6: Config parser

Right now, we have a very crude configuration parser: we use `getEnv` to read two environment variables, `SLACK_TZ_APP_TOKEN` and `SLACK_TZ_BOT_TOKEN`.

We should implement a proper config parser. The app should look for a file named `config.dhall` (see: [dhall](https://hackage.haskell.org/package/dhall/)), and parse the options below.

* `appToken`/`SLACK_TZ_APP_TOKEN`: required
* `botToken`/`SLACK_TZ_BOT_TOKEN`: required
* `maxRetries`/`SLACK_TZ_MAX_RETRIES`: optional, defaults to 3. This will be used in #5.
* `cacheUsersInfo`/`SLACK_TZ_CACHE_USERS_INFO`: optional, defaults to 1 hour. This will be used in #7.
* `cacheConversationMembers`/`SLACK_TZ_CACHE_CONVERSATION_MEMBERS`: optional, defaults to 1 minute. This will be used in #7.

Environment variables take precedence over the config file. If all the required options are specified as environment variables, then a config file does not need to exist on disk.

Tip: we can use the `o-clock` package to express and parse units of time:

```hs
> unitsP @Second "300s"
Just (300s)

> unitsP @Second "5m"
Just (300s)
```

# Acceptance criteria

* The app parses the environment variables above.
* The user can also choose to write them in a config file.
* Env vars take precedence over the config file.

YuriRomanowski:
## Description

Problem: Currently, bot accepts no configuration files and directly
reads tokens from environment variables.

Solution: Add configuration loading stage, load configuration from
environment variables and configuration files, with default settings.

## Related issue(s)

Fixed #6 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


alexd1971 (commented):
Maybe I'm wrong, but I think that all code from module `TzBot.Config` could be replaced with one line:
```haskell
loadYamlSettings ["config/config.yaml", "defaults.yaml"] [] useEnv
```

YuriRomanowski:
> Maybe I'm wrong, but I think that all code from module TzBot.Config could be replaced with one line:

In the original issue, there was a requirement that if all the data is available in the environment variables then we don't need any config files at all, and for this we should be able to parse envvars without reading config, and we can't achieve this with `loadYamlSettings`.

alexd1971:
> In the original issue, there was a requirement that if all the data is available in the environment variables then we don't need any config files at all, for this we should be able to parse envvars without reading config, and we can't achieve this with only `loadYamlSettings`.

I guess the following solution satisfies the mentioned requirement, but seems to be much simpler:
```haskell
{-# LANGUAGE LambdaCase #-}
module Main where

import Data.Yaml.Config
import Data.Yaml
import System.Directory

configFile = "config.yaml"
defaults = "defaults.yaml"

main :: IO ()
main = do
  configFiles <- doesFileExist configFile >>= \case
    True -> pure [configFile, defaults]
    False -> pure [defaults]
  config <- loadYamlSettings configFiles [] useEnv :: IO Value
  print config
```

YuriRomanowski:
>>> I guess the following solution satisfies the mentioned requirement, but seems to be much simpler:

For the case when the user file is absent, we need to write all the envvars in the default file. But if it is present, its contents have precedence over the default file, so we have to rewrite all the envvars there again, because we want envvars to have the highest priority. So our config files will be quite cluttered with environment variable names. The approach used in this PR results in much more clean configuration files.

src/TzBot/Config/Types.hs
```diff
@@ -0,0 +1,71 @@
...
+newtype BotToken = BotToken
+  { unBotToken :: Text }
+  deriving stock (Eq, Show, Generic, Data)
+  deriving newtype (FromJSON, ToJSON, NFData)
+
+data ConfigOrigin =
+    COInterm -- ^ Fields are optional and contain info about the envvar names.
+  | COTotal  -- ^ All fields are present.
+
+type family ConfigField (k :: ConfigOrigin) a where
```
alexd1971, line 36: What is the purpose of this type family? Why we cannot use just `Maybe a` and `a` in `Config` type?

alexd1971 (approved):
Approved. But I would suggest to ask someone else to review this PR. IMO this implementation is overcomplicated for simple configuration consisting only 3 parameters. But maybe this implementation can make our life easier in future. Unfortunaly I can't see that for now.

src/TzBot/Config/Types.hs
YuriRomanowski: Our final configuration shouldn't contain any optional fields, while the configuration parsed from file may miss some fields. But this type family allows us not to create two different datatype for these two cases.

PR #25, opened 2022-12-29
[#7] Cache slack data

### LINKED ISSUE #7: Cache slack data

We have [2 methods](https://github.com/serokell/tzbot/blob/main/src/TzBot/Slack/WebAPI/Class.hs) for retrieving data from Slack's Web API, `getUser` and `getChannelMembers`. Both endpoints are subject to Slack's [rate limits](https://api.slack.com/docs/rate-limits), so we should refrain from making needless calls to avoid hitting these limits.

To this end, we should cache the data retrieved from these 2 methods.

To keep a clean separation of concerns, we should implement the caching as a monad transformer that has an instance of `WebApi` and can be stacked on top of `WebAPIM`.

Bonus points: we should also make sure that cache entries don't all expire at the same time (otherwise, we risk having to refresh hundreds of entries at once and, thus, hitting the rate limits). Expiry times should be somewhat randomized. Users' info should be cached for about 1h ~ 1h30m and channels' membership info for about 1~2 minutes.

# Acceptance criteria

* The data retrieved from `getUser` and `getChannelMembers` is cached
* Bonus: each cache entry's expiry time is random (within some bounds)

YuriRomanowski:
## Description

Problem: The bot uses two methods, `getUser` and `getChannelMembers`, which are subjects to Slack's rate limits, and we need to avoid unneeded calls to those methods.

Solution: Add caching for user info and channel members.

## Related issue(s)

Fixed #7

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/TzBot/BotMain.hs
+++ b/src/TzBot/BotMain.hs
@@ -6,6 +6,7 @@ module TzBot.BotMain where
 
 import Universum
 
+import Control.Monad.Managed (managed, runManaged)
 import Data.Aeson.Types (FromJSON(parseJSON), parseEither)
 import Data.Map qualified as M
 import Data.Text qualified as T
@@ -16,9 +17,11 @@ import Slacker
   pattern EventValue, runSocketMode, setApiToken, setAppToken, setOnException)
 import System.Environment (getArgs)
 
-import TzBot.Config (AppLevelToken(..), BotToken(..), Config(cAppToken, cBotToken), readConfig)
-import TzBot.ProcessEvent (processEvent)
-import TzBot.RunMonad (BotState(BotState), log', runBotM)
+import TzBot.Cache
+import TzBot.Config (AppLevelToken(..), BotToken(..), Config(..), readConfig)
+import TzBot.ProcessEvent
+  (processMemberJoinedChannel, processMemberLeftChannel, processMessageEvent)
+import TzBot.RunMonad (BotState(..), log', runBotM)
 
 {- |
 
@@ -33,25 +36,32 @@ main :: IO ()
 main = do
   -- TODO: add optparse
   (configFilePath :: Maybe FilePath) <- safeHead <$> getArgs
-  config <- readConfig configFilePath
+  bsConfig@Config{..} <- readConfig configFilePath
 
-  let appLevelToken = cAppToken config
-      botToken = cBotToken config
   let sCfg = defaultSlackConfig
-            & setApiToken (unBotToken botToken)
-            & setAppToken (unAppLevelToken appLevelToken)
+            & setApiToken (unBotToken cBotToken)
+            & setAppToken (unAppLevelToken cAppToken)
             & setOnException handleThreadExceptionSensibly -- auto-handle disconnects
 
-  manager <- newManager tlsManagerSettings
-  refSetMapIORef <- newIORef M.empty
-  let bState = BotState config manager refSetMapIORef
-
-  runSocketMode sCfg (handler bState) -- auto-acknowledge received messages
+  bsManager <- newManager tlsManagerSettings
+  bsMessagesReferences <- newIORef M.empty
+  runManaged $ do
+    bsUserInfoCache <-
+      managed $ withRandomizedCacheDefault cCacheUsersInfo
+    bsConversationMembersCache <-
+      managed $ withRandomizedCacheDefault cCacheConversationMembers
+    liftIO $ runSocketMode sCfg $ handler BotState {..}
 
 handler :: BotState -> SlackConfig -> SocketModeEvent -> IO ()
 handler bState _cfg = \e -> do
   case e of
     EventValue "message" evtRaw -> case parseEither parseJSON evtRaw of
       Left err -> log' $ T.pack err
-      Right evt -> void . runBotM bState $ processEvent evt
+      Right evt -> void . runBotM bState $ processMessageEvent evt
+    EventValue "member_joined_channel" evtRaw -> case parseEither parseJSON evtRaw of
+      Left err -> log' $ T.pack err
+      Right evt -> void . runBotM bState $ processMemberJoinedChannel evt
+    EventValue "member_left_channel" evtRaw -> case parseEither parseJSON evtRaw of
+      Left err -> log' $ T.pack err
+      Right evt -> void . runBotM bState $ processMemberLeftChannel evt
     _ -> pure ()
--- a/src/TzBot/Cache.hs
+++ b/src/TzBot/Cache.hs
@@ -1,0 +1,139 @@
+-- SPDX-FileCopyrightText: 2022 Serokell <https://serokell.io/>
+--
+-- SPDX-License-Identifier: MPL-2.0
+
+module TzBot.Cache
+  ( -- * Types
+    RandomizedCache
+
+    -- * Cache creation
+  , withRandomizedCache
+  , withRandomizedCacheDefault
+
+    -- * Altering cache
+  , insertRandomized
+  , fetchWithCacheRandomized
+  , update
+  ) where
+
+import Universum
+
+import Control.Concurrent.Async.Lifted (withAsync)
+import Data.Cache (Cache)
+import Data.Cache qualified as Cache
+import Data.Cache.Internal qualified as CacheI
+import Data.HashMap.Strict qualified as HM
+import System.Clock (TimeSpec)
+import Time (Hour, KnownDivRat, Nanosecond, Time(..), hour, threadDelay, toUnit)
+
+import TzBot.Util (randomTimeSpec, timeToTimespec, (+-))
+
+-- | This datatype uses `Cache` datatype inside, but also
+--   provides a possibility to slightly vary expiration
+--   time randomly around the configured mean value.
+--   Use it when you don't want a bunch of cached data
+--   to expire at once.
+--
+--   Automatic periodical cleaning is also included.
+data RandomizedCache k v = RandomizedCache
+  { rcCache  :: Cache k v
+  , rcExpiry :: TimeSpec
+  , rcExpiryRandomAmplitude :: TimeSpec
+  }
+
+-- | Like `withRandomizedCache` but with reasonable defaults.
+-- Expiry random amplitude is 15% of the expiry time,
+-- default cleaning period of 24 hours is used.
+withRandomizedCacheDefault
+  :: ( KnownDivRat u Nanosecond
+     , Eq k, Hashable k
+     )
+  => Time u -- ^ Expiration time
+  -> (RandomizedCache k v -> IO a) -- ^ Action that uses the cache
+  -> IO a
+withRandomizedCacheDefault expiry =
+  withRandomizedCache expiry (get15PerCent expiry) Nothing
+  where
+    get15PerCent :: Time k -> Time k
+    get15PerCent (Time r) = Time $ r * 0.15
+
+-- | Create randomized cache and run the passed action with it.
+-- Raises `error` on incorrect use.
+withRandomizedCache
+  :: ( KnownDivRat u Nanosecond
+     , Eq k, Hashable k
+     , HasCallStack
+     )
+  => Time u -- ^ Expiration time
+  -> Time u -- ^ Expiration time amplitude (should be less than expiration time)
+  -> Maybe (Time Hour) -- ^ How frequently to purge all expired items
+  -> (RandomizedCache k v -> IO a) -- ^ Action that uses the cache
+  -> IO a
+withRandomizedCache
+    (timeToTimespec -> rcExpiry)
+    (timeToTimespec -> rcExpiryRandomAmplitude)
+    cleaningPeriod
+    action
+  = do
+  when (rcExpiry - rcExpiryRandomAmplitude <= 0) $
+    error "Expiry random amplitude should be lower than the expiry"
+  rcCache <- Cache.newCache $ Just rcExpiry
+  withAsync
+    (cleaningThread (toUnit <$> cleaningPeriod) rcCache)
+    (\_ -> action RandomizedCache {..})
+
+defaultCleaningPeriod :: Time Hour
+defaultCleaningPeriod = hour 24
+
+cleaningThread :: (Eq k, Hashable k) => Maybe (Time Hour) -> Cache k v -> IO ()
+cleaningThread mbCleaningPeriod cache = forever $ do
+  Time.threadDelay cleaningPeriod
+  Cache.purgeExpired cache
+  where
+    cleaningPeriod = fromMaybe defaultCleaningPeriod mbCleaningPeriod
+
+-- | Generate a random expiry time and insert a key/value pair into
+-- the cache with that expiry time.
+insertRandomized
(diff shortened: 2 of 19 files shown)


YuriRomanowski:
Guys how do you think, do we need some tests here for the actual events processing and caching?

alexd1971:
Probably it would be better to rebase this branch onto main before review.

src/TzBot/Slack.hs
```diff
@@ -37,13 +38,24 @@ getUser userId = do
   token <- getBotToken
   usersInfo token userId >>= endpointFailed "users.info"
 
+-- | Get a user's info using cache.
+getUserCached :: UserId -> BotM User
+getUserCached userId =
+  asks bsUserInfoCache
+    >>= Cache.fetchWithCacheRandomized userId getUser
```
Martoon-00, line 45: The issue description mentioned reusing the `WebAPI` typeclass, and as a way to handle effects provide two monads - one that instantiates `WebAPI` with plain `getUser` and another that uses cached versions.

But I remember that this typeclass is now gone, so I dunno what to do :sweat_smile: 

Given this use case, would it make sense for you to bring that typeclass back? Maybe without some of its methods if they don't make much sense today.

src/TzBot/Util.hs
```diff
@@ -26,3 +29,20 @@ thTrickYaml :: forall a. (FromJSON a) => ByteString -> Trick a
...
   Right c -> ([|id|], c)
+
+timeToTimespec :: KnownDivRat k Nanosecond => Time k -> TimeSpec
+timeToTimespec = fromNanoSecs . floorRat . toUnit @Nanosecond
+
+timespecToTime :: KnownDivRat Nanosecond k => TimeSpec -> Time k
+timespecToTime = toUnit . ns . fromIntegral . toNanoSecs
+
+{-
+>>> let Just time = unitsP "30m" :: Maybe (Time Minute) in (timespecToTime $ timeToTimespec time :: Time Minute) == time
```
Martoon-00, line 40: I once wondered, whether our rule about 80 chars per line touches the comments.

I think, one major reason for this practice is that in narrow editors (e.g. Vim in terminal or Emacs split in several panes) long lines look ugly.

Moreover, using `>>>` multiple times in consequence seems to be a valid syntax, and I don't see reasons not to do so here, given that this is a `let .. in`.

src/TzBot/BotMain.hs
```diff
@@ -33,25 +35,29 @@ main :: IO ()
...
-  let wState = BotState wCfg manager refSetMapIORef
+  bsManager <- newManager tlsManagerSettings
+  bsMessagesReferences <- newIORef M.empty
+  let get15PerCent :: Time k -> Time k
+      get15PerCent (Time r) = Time $ r * 0.15
+  bsUserInfoCache <-
+    newRandomizedCacheThrow cCacheUsersInfo (get15PerCent cCacheUsersInfo) Nothing
+  bsConversationMembersCache <-
+    newRandomizedCacheThrow cCacheConversationMembers
+      (get15PerCent cCacheConversationMembers) Nothing
```
Martoon-00, line 58: Hmm, I'm still skeptical about the need in expiration for these two cases, but I only now noticed that the ticket description says to do so; let's go so.

src/TzBot/Cache.hs
```diff
@@ -0,0 +1,110 @@
...
+
+defaultCleaningPeriod :: Time Hour
+defaultCleaningPeriod = hour 24
+
+cleaningThread :: (Eq k, Hashable k) => Maybe (Time Hour) -> Cache k v -> IO ()
+cleaningThread mbCleaningPeriod cache = forever $ do
+  Time.threadDelay cleaningPeriod
+  Cache.purgeExpired cache
+  where
+    cleaningPeriod = fromMaybe defaultCleaningPeriod mbCleaningPeriod
```
Martoon-00, line 84: Do I understand it correctly, that with this scheme we will load fresh data only once per hour by default? And if e.g. a new user joins a channel in Slack, we may notice that only in an hour?

Martoon-00 (commented):
Neat code, and very good work :+1:

src/TzBot/Cache.hs
```diff
@@ -0,0 +1,110 @@
...
+  -> IO (Either String $ RandomizedCache k v)
+newRandomizedCache
+    (timeToTimespec -> rcExpiry)
+    (timeToTimespec -> rcExpiryRandomAmplitude)
+    cleaningPeriod
+  = runExceptT $ do
+  when (rcExpiry - rcExpiryRandomAmplitude <= 0) $
+    throwError "Expiry random amplitude should be lower than the expiry"
+  rcCache <- lift $ Cache.newCache $ Just rcExpiry
+  lift $ forkIO $ cleaningThread (toUnit <$> cleaningPeriod) rcCache
```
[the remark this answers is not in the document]
YuriRomanowski: I thought that here it is acceptable because we expect this thread to work until the program ends?

src/TzBot/Slack.hs
YuriRomanowski: I decided to delete `WebAPI` because for that moment it had only one instance and only made the modules structure more complicated, and later we can bring it back (maybe not `WebAPI` exactly, maybe other typeclasses), maybe for testing, but for now it's more convenient to use just IO like monad without any mtl style.

src/TzBot/Cache.hs
```diff
@@ -0,0 +1,110 @@
...
+import Control.Concurrent
+import Control.Monad.Except
+import Data.Cache (Cache)
+import Data.Cache qualified as Cache
+import System.Clock
+import Time
+
+import TzBot.Util (randomTimeSpec, timeToTimespec, (+-))
+
+-- | This datatype uses `Cache` datatype inside, but also
```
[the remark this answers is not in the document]
YuriRomanowski: Their meaning is absolutely [the same](https://haskell-haddock.readthedocs.io/en/latest/markup.html#hyperlinked-identifiers)

src/TzBot/Cache.hs
YuriRomanowski: You are right, if the channel members caching expiration time is one hour, we will notice the new user only after that hour is gone.
I think, we can combine this scheme with listening for the events [member_joined_channel](https://api.slack.com/events/member_joined_channel) and [member_left_channel](https://api.slack.com/events/member_left_channel). The former allows to get the majority of the required information at once, and the latter allows to keep that information in a valid state.

src/TzBot/Cache.hs
```diff
@@ -0,0 +1,110 @@
...
+  -> Time u -- ^ Expiration time amplitude (should be less than expiration time)
+  -> Maybe (Time Hour) -- ^ How frequently to purge all expired items
+  -> IO (Either String $ RandomizedCache k v)
+newRandomizedCache
+    (timeToTimespec -> rcExpiry)
+    (timeToTimespec -> rcExpiryRandomAmplitude)
+    cleaningPeriod
+  = runExceptT $ do
+  when (rcExpiry - rcExpiryRandomAmplitude <= 0) $
+    throwError "Expiry random amplitude should be lower than the expiry"
```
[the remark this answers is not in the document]
YuriRomanowski: I intentionally created two different functions: returning `Either` and plain IO function, so the user can use whatever he wants.

src/TzBot/Cache.hs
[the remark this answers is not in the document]
YuriRomanowski: You are right, let's use `withAsync` here, it would be helpful if the cache is being used not during all the program work, but for a specific code block (although it's not the case currently).

src/TzBot/Util.hs
YuriRomanowski: OK, let's split it into two lines

src/TzBot/ProcessEvent.hs
```diff
@@ -69,25 +69,23 @@ processEvent :: MessageEvent -> BotM ()
...
+    usersInChannelIds <- getChannelMembersCached channelId
     let ephemeralTemplate = NE.map (renderEphemeralMessageTemplate now sender) timeRefsUtc
 
     whenJust (renderErrorsForSender ephemeralTemplate) $ \errorsMsg ->
       sendAction errorsMsg (uId sender)
 
     let notBotAndSameTimeZone u = not (uIsBot u) && uTz u /= uTz sender
         notSender userId = userId /= uId sender
-    usersToNotify <- filter notBotAndSameTimeZone <$> (mapM getUser $ filter notSender usersInChannelIds)
+    usersToNotify <- filter notBotAndSameTimeZone <$> (mapM getUserCached $ filter notSender usersInChannelIds)
```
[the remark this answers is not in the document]
YuriRomanowski: What you say seems good to be a different issue, how about creating some new tickets for this?

src/TzBot/Cache.hs
```diff
@@ -0,0 +1,110 @@
...
+import Data.Cache qualified as Cache
+import System.Clock
+import Time
+
+import TzBot.Util (randomTimeSpec, timeToTimespec, (+-))
+
+-- | This datatype uses `Cache` datatype inside, but also
+--   provides a possibility to slightly vary expiration
+--   time randomly around the configured mean value.
+--   Use it when you don't want a barch of cached data
```
YuriRomanowski, line 32: Hm, what is "barch"??? Google translate says it means "respect" in Welch, I have no idea how this word could appear here :laughing:

src/TzBot/Cache.hs
```diff
@@ -0,0 +1,110 @@
...
+--   to expire at once.
+--
+--   Automatic periodical cleaning is also included.
```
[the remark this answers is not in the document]
YuriRomanowski: Partly it does provide, when we lookup an entry and it turns out to be expired, it just gets deleted and the lookup just returns `Nothing`, but we also want to clean it periodically not to keep a lot of expired data. And for our purposes we don't need to do it very frequently because the number of users is not very big. But this may play a bigger role later when we use this caching for keeping messages.

src/TzBot/Cache.hs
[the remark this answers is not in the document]
Martoon-00: Exactly. E.g. if one day we add integrational tests, auto-stopping would be handy.

src/TzBot/Cache.hs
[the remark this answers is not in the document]
Martoon-00: Oh I see, TIL

src/TzBot/Cache.hs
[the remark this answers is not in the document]
Martoon-00: Well, if we end up duplicating every function with preconditions this way, that would soon become annoying.

Please follow [this guide](https://www.notion.so/serokell/Guidelines-for-Exception-Handling-in-Cardano-SL-fb34acf10f1a4f4fadbf9b44ad81c44b). I would not insist on applying it completely, but at least on choosing between `error` / `ExceptT` / custom exception - it should be followed.

src/TzBot/ProcessEvent.hs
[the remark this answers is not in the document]
Martoon-00: Agreed. Created https://github.com/serokell/tzbot/issues/35 for this.

src/TzBot/Slack.hs
Martoon-00: > for now it's more convenient to use just IO-like monad without any mtl style.

Okay, that sounds convincing.

stack.yaml
```diff
@@ -7,6 +7,8 @@ resolver: lts-19.23
 packages:
 - .
 
+system-ghc: true
+
```
Martoon-00, line 11: Seems like something foreign leaked into here

src/TzBot/Cache.hs
```diff
@@ -7,22 +7,27 @@ module TzBot.Cache
...
   , fetchWithCacheRandomized
+  , update
   ) where
 
 import Universum
 
-import Control.Concurrent (forkIO)
-import Control.Monad.Except (MonadError(throwError))
+import Control.Concurrent.Async.Lifted (withAsync)
+import Control.Monad.Managed (Managed, managed)
```
Martoon-00, line 24: Neat, I didn't know about this module and have only used `ContT`-based bicycle

src/TzBot/Cache.hs
```diff
@@ -108,3 +147,17 @@ fetchWithCacheRandomized key fetchAction cache = do
...
+  => k
+  -> (v -> v)
+  -> RandomizedCache k v
+  -> m ()
+update key valFunc RandomizedCache {..} = do
+  atomically $ modifyTVar' (CacheI.container rcCache) $ \hm ->
+    HM.adjust itemFunc key hm
+  where
+    itemFunc :: CacheI.CacheItem v -> CacheI.CacheItem v
+    itemFunc ci = ci { CacheI.item = valFunc $ CacheI.item ci }
```
Martoon-00, line 163: Just meeh, looking at this I guess the library choice turned out to be too sad :unamused: 

But it's written and it works, so should be good to go. I would just consider trying to switch to something smarter in the nearest time, WDYT?

src/TzBot/Cache.hs
Martoon-00: That sounds (and now looks) very good :+1:

Martoon-00 (commented):
I much like it, starting from some stylistic cleanups, use of logging (even if it is raw - better now than to insert it everywhere later), end finishing with handling channels' participants changes, that feels like an apparent must-have and good that we already implemented it.

I'm leaving a few more small comments, and that should be it.

stack.yaml
YuriRomanowski: Oh, my...

src/TzBot/Cache.hs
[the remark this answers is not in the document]
YuriRomanowski: Thank you, let's apply this guide

src/TzBot/Cache.hs
YuriRomanowski: I fully share your feelings, the library really doesn't meet our requirements in any reasonable extent, it's now used as a template for our own caching.

YuriRomanowski: I decided to try to get something working from what we have, but yes, it's worth searching for something else. Going to do it soon

YuriRomanowski: It seems that it's a common pattern to take some library and use it as a base of our own implementation for this purpose, but still we can use some other data structures (maybe `psqueues` should fit well), but I would say the the current wrapper is also OK.

src/TzBot/Cache.hs
```diff
@@ -0,0 +1,138 @@
...
+  withRandomizedCache expiry (get15PerCent expiry) Nothing
+  where
+    get15PerCent :: Time k -> Time k
+    get15PerCent (Time r) = Time $ r * 0.15
+
+-- | Create randomized cache and run the passed action with it.
+-- Raises `error` on incorrect use.
+withRandomizedCache
+  :: ( KnownDivRat u Nanosecond
+     , Eq k, Hashable k
```
Martoon-00, line 64: Minor: adding `HasCallStack` could also be helpful, just in case.

Martoon-00 (approved):
Looks very good, thank you for the work :+1:

PR #26, opened 2022-12-29
[#8] Handle Slack pagination

### LINKED ISSUE #8: Pagination of the `conversation.members` endpoint

At the moment, we call Slack's [`conversation.members` endpoint](https://api.slack.com/methods/conversations.members/) to retrieve a list with the IDs of the users in a given channel.

We expect the response body to be a JSON object with two keys, `ok` and `members` (see [here](https://github.com/serokell/tzbot/blob/16e138e7496dc69068c1f3ac5996f3cd22d121ca/src/TzBot/Slack/WebAPI/API.hs#L38-L42) and [here](https://github.com/serokell/tzbot/blob/16e138e7496dc69068c1f3ac5996f3cd22d121ca/src/TzBot/Slack/WebAPI/API.hs#L55-L80))

However, this endpoint may return paginated data.

The response contains a third key, `response_metadata`, with the cursor for retrieving the next page (by passing it as a `?cursor=` query parameter). E.g.:

```json
{
    "ok": true,
    "members": [
        "U02N4G9NYR4",
        "U02N85E78QM"
    ],
    "response_metadata": {
        "next_cursor": ""
    }
}
```

When the last page is reached, `next_cursor` is an empty string.

We should make sure our servant API supports that and our client implementation (`TzBot.Slack.WebAPI.Impl`) follows those cursors.

Relevant docs:
* `conversation.members` endpoint: https://api.slack.com/methods/conversations.members/test
* Paginating through collections: https://api.slack.com/docs/pagination

# Acceptance criteria

* The `getChannelMembers` method follows the cursors of Slack's `conversatio.members` endpoint

YuriRomanowski:
## Description

Problem: Some Slack methods are paginated and hence can return not all objects if the limit turns out to be less than total amount of the objects.

Solution: Handle pagination by cursor following, fetching all available objects.

## Related issue(s)

Fixed #8 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/TzBot/Slack.hs
+++ b/src/TzBot/Slack.hs
@@ -23,6 +23,7 @@ import Control.Concurrent (threadDelay)
 import Control.Monad.Except (throwError)
 import Data.Aeson (Value)
 import Data.ByteString.UTF8 (toString)
+import Data.DList qualified as DList
 import Data.Sequence qualified as Seq
 import Data.Text qualified as T
 import Data.Text.Encoding qualified as T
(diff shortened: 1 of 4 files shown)


src/TzBot/Slack.hs
```diff
@@ -67,25 +68,39 @@ getBotToken = do
...
+getPaginatedObjects :: Text -> (Maybe Cursor -> BotM $ SlackResponse $ SlackContents key [a]) -> BotM [a]
+getPaginatedObjects endpoint action = DList.toList <$> go mempty Nothing
+  where
+    go acc mbCursor = do
+      eithRes <- action mbCursor
+      res <- handleSlackError endpoint eithRes
+      let newAcc = acc <> DList.fromList (scContents res)
+      case scCursor res of
+        Nothing -> pure newAcc
+        newMbCursor@(Just _) -> go newAcc newMbCursor
```
Martoon-00, line 90: That's subjective, but I find avoiding wildcard patterns this way pretty nice.

Martoon-00 (approved):
Looks good :+1:

PR #28, opened 2023-01-06
Handle 429 Too Many Requests

### LINKED ISSUE #5: Handle "429 Too Many Requests"

Sometimes, we may hit Slack's [rate limits](https://api.slack.com/docs/rate-limits) and Slack's Web API will start responding with `429 Too Many Requests`. When this happens, the response should have a `Retry-After` header indicating how many seconds we should wait before retrying the request.

You can find the current implementation of the Web API client in `TzBot.Slack.WebAPI.Impl`.

We should give up after 3 attempts (we can make this number configurable later).

# Acceptance criteria

* We gracefully handle Slack's rate limits.
* The app doesn't crash when we get a 429.
* We wait the given number of seconds before retrying the request.

alexd1971:
## Description

Problem: Sometimes slack can respond with status 429 Too Many Requests.
We have to handle this and take a try to request the slack api again
after the delay set in Retry-After header.

Solution: Wrap enent handler with function which detects the response
with status code 429, extracts the delay from headers and runs again the
event handler after the delay. The number of attempts is defined in bot
configuration.

## Related issue(s)

Fixed #5 

- Tests
  - [X] If I added new functionality, I added tests covering it.
        silently reappearing again.

- Documentation
  - [X] I checked whether I should update the docs and did so if necessary:
    - Haddock

- [X] My commits comply with [the policy used in Serokell](https://www.notion.so/serokell/Where-and-how-to-commit-your-work-58f8973a4b3142c8abbd2e6fd5b3a08e).
- [X] My code complies with the [style guide](../tree/master/docs/code-style.md).

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -51,8 +51,6 @@ library:
   - tztime
   - validation
   - yaml
-  - uri-bytestring
-  - uri-bytestring-aeson
   - utf8-string
 
 executables:
--- a/package.yaml
+++ b/package.yaml
@@ -28,6 +28,7 @@ library:
   - formatting
   - http-client
   - http-client-tls
+  - http-types
   - lifted-async
   - megaparsec
   - monad-control
@@ -37,6 +38,7 @@ library:
   - servant-auth
   - servant-auth-client
   - servant-client
+  - servant-client-core
   - servant-server
   - slacker
   - string-conversions
@@ -49,6 +51,9 @@ library:
   - tztime
   - validation
   - yaml
+  - uri-bytestring
+  - uri-bytestring-aeson
+  - utf8-string
 
 executables:
   tzbot-exe:
@@ -82,6 +87,11 @@ tests:
     - tasty-hunit
     - time
     - tztime
+    - QuickCheck
+    - http-types
+    - servant-client-core
+    - tasty-quickcheck
+    - time
     - universum
 
 ghc-options:
--- a/src/TzBot/BotMain.hs
+++ b/src/TzBot/BotMain.hs
@@ -6,7 +6,7 @@ module TzBot.BotMain where
 
 import Universum
 
-import Data.Aeson.Types
+import Data.Aeson.Types (FromJSON(parseJSON), parseEither)
 import Data.Map qualified as M
 import Data.Text qualified as T
 import Network.HTTP.Client (newManager)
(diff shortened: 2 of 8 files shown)


YuriRomanowski:
When I added the config including `maxRetries`, unfortunately I didn't think about validating it, but now negative values can cause infinite loops, so could you please add its validation during the config loading?

alexd1971:
> When I added the config including `maxRetries`, unfortunately I didn't think about validating it, but now negative values can cause infinite loops, so could you please add its validation during the config loading?

Do we actually need this configuration parameter? It value can't vary in broad limits. Possibly it can be 1 or 2 retries. Maybe we could hard set it at 2 retries?

alexd1971:
> but now negative values can cause infinite loops

I can make a fix that handler will retry only if `maxRetries` is greater than zero. Does it make sense?

YuriRomanowski:
> > but now negative values can cause infinite loops
> 
> I can make a fix that handler will retry only if `maxRetries` is greater than zero. Does it make sense?

Yes, this is also possible, let's do that.

YuriRomanowski:
> > When I added the config including `maxRetries`, unfortunately I didn't think about validating it, but now negative values can cause infinite loops, so could you please add its validation during the config loading?
> 
> Do we actually need this configuration parameter? It value can't vary in broad limits. Possibly it can be 1 or 2 retries. Maybe we could hard set it at 2 retries?

I don't see a big difference between configured or hard-coded parameter now, but the former option is just more flexible, why not?

YuriRomanowski:
Also another big concern: in the PR we just wrap the whole event handler action in retry function, but the handler can be rather complex and contain a lot of API calls, especially I'm concerned about really dirty ones like messages posting. Imagine that we have to post several amount of ephemeral messages (say, 10). With what we have now in this PR, this can easily lead to a situation that we will succeed to post five of them, but fail on 6th message, and then we will retry from the very beginning, posting those 5 messages again, and this is obviously undesirable.

I didn't want this myself, but probably we have no choice: it seems we have to wrap every single API call with retry function. Then, by default triggering these API calls will fail all the handler, but we'll be able to use `catchError` if it's required.

alexd1971:
> it seems we have to wrap every single API call with retry function

Ok. I'll think how to implement this

YuriRomanowski (approved):

PR #37, opened 2023-01-17
[#33] Add logging

### LINKED ISSUE #33: Add logging

Currently in fact there is no normal logging, just slightly modified `putStrLn` is used. I think we should introduce some logging library that meets all the requirements of good logging system.

# Acceptance criteria
* Choose the good logging library
* Replace `print`s with this logging

YuriRomanowski:
## Description
Problem: Currently, the bot doesn't have any logging infrastructure, logging is performed only by slightly modified `putStrLn` function.

Solution: Add use of the logging library `katip`. Minimal log level can be configured, only logging to stdout is supported.

Also fixed: ignore messages from other bots (and from ourselves)

## Related issue(s)

Fixed #33 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/TzBot/Logger.hs
+++ b/src/TzBot/Logger.hs
@@ -1,0 +1,75 @@
+-- SPDX-FileCopyrightText: 2022 Serokell <https://serokell.io/>
+--
+-- SPDX-License-Identifier: MPL-2.0
+
+module TzBot.Logger
+  ( module TzBot.Logger
+  , KatipContext (..)
+  , KatipContextT
+  , katipAddNamespace
+  , katipAddContext
+  ) where
+
+import Universum
+
+import Data.Aeson (KeyValue((.=)), ToJSON(..), object)
+import Katip
+
+import TzBot.Slack.API (MessageId(..))
+
+logSugar_ :: (KatipContext m, HasCallStack) => Severity -> Text -> m ()
+logSugar_ sev = logLocM sev . logStr
+
+logInfo, logWarn, logDebug, logError :: (KatipContext m, HasCallStack) => Text -> m ()
+logInfo t = withFrozenCallStack $ logSugar_ InfoS t
+logWarn t = withFrozenCallStack $ logSugar_ WarningS t
+logDebug t = withFrozenCallStack $ logSugar_ DebugS t
+logError t = withFrozenCallStack $ logSugar_ ErrorS t
+
+withLogger
+  :: Severity
+  -> ((Namespace, LogContexts, LogEnv) -> IO a)
+  -> IO a
+withLogger logLevel action = do
+  handleScribe <- mkHandleScribe ColorIfTerminal stdout (permitItem logLevel) V2
+  let namespace = "main"
+      initialContext = mempty
+      environment = "general"
+      mainNamespace = "TzBot"
+      scribeName = "stdout"
+      mkLogEnv =
+        initLogEnv mainNamespace environment >>=
+        registerScribe scribeName handleScribe defaultScribeSettings
+  bracket mkLogEnv closeScribes $ \logEnv ->
+    action (namespace, initialContext, logEnv)
+
+katipAddNamespaceText :: (KatipContext m) => Text -> m a -> m a
+katipAddNamespaceText txt = katipAddNamespace $ Namespace [txt]
+
+-- contexts
+type EventType = Text
+type EventId = Text
+data EventContext = EventContext EventType EventId
+
+instance ToJSON EventContext where
+  toJSON (EventContext t cId) =
+    object
+      [ "event_type" .= t
+      , "event_id" .= cId
+      ]
+
+instance ToObject EventContext
+
+instance LogItem EventContext where
+  payloadKeys _verb _a = AllKeys
+
+--
+newtype MessageContext = MessageContext MessageId
+
+instance ToJSON MessageContext where
+  toJSON (MessageContext msgId) = object ["message_id" .= msgId]
+
+instance ToObject MessageContext
+
+instance LogItem MessageContext where
+  payloadKeys _verb _a = AllKeys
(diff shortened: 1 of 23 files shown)


YuriRomanowski:
I tried to use `katip`, but I'm going to look at other libs more closely, maybe later I'll change it

src/TzBot/Slack.hs
```diff
@@ -232,48 +236,59 @@ usersInfo
...
+                      return $ Left err -- if there is no Retry-After header don't handle 429 response
+                    Just retryAfter -> do
+                      let delay = sec $ realToFrac retryAfter
+                      $(logTM) `error_`
+                        [int||Retrying again in #{retryAfter}, #{attempts-1} attempts remaining|]
+                      threadDelay delay
+                      modify (subtract 1)
+                      handleTooManyRequests botAction
+                else do
+                  $(logTM) `error_` "Attempts amount exceeded, stopping"
```
Martoon-00, line 291: Really good to have this logged :+1: 

In case we get several overlapping retrying routines: will these logs be accompanied by some unique identifiers in the log namespace, so that we could distinguish Bot's reactions on different messages?

src/TzBot/Util.hs
```diff
@@ -156,3 +160,22 @@ instance (FromJSON a) => FromJSON (WithUnknown a) where
...
+postfixFields :: LensRules
+postfixFields = lensRules & lensField .~ mappingNamer (\n -> [n ++ "L"])
+
+----
+-- not present in mtl-2.2.2
+tryError :: MonadError e m => m a -> m (Either e a)
+tryError action = (Right <$> action) `catchError` (pure . Left)
+
+catchAllErrors
+  :: (MonadError e m, MonadBaseControl IO m)
```
Martoon-00, line 173: I guess you use `MonadBaseControl` because `ExceptT` layer does not pass `MonadCatch`?

Hmm, there is a problem with this: we generally try to use exception utilities that don't catch asynchronous exceptions by default. E.g. `safe-exception` package, `unliftio` package with its `UnliftIO.Exception` module, and `universum` do this, while classic `Control.Exceptions`, `mtl`-style exceptions, and `Control.Exception.Lifted` don't.

Maybe there would be other easy way to workaround that, without using other `try`?

Martoon-00 (commented):
Very nice to see the key events accompanied by logs, and generally logs are now looking quite neat and with colors :yum: 

I'm leaving a few comments.

src/TzBot/Slack.hs
YuriRomanowski: In the `ProcessEvents.hs` module I wrapped every handler with `katipAddContext`, and for messages I additionally  wrapped the corresponding handler with `katipAddContext` and `katipAddNamespace` functions, and these log tags are propagated to the separate threads, so I think we will be able to distinguish. But if different routines are for the same message, the only difference will be thread id, so here we cannot distinguish (but maybe that's OK).

Martoon-00: Ah right, we have thread ids. I guess it should be enough, really

src/TzBot/ProcessEvents.hs
```diff
@@ -47,51 +47,64 @@ the \"report\" dialog. And instead \"report\" dialog contains
...
-handler wstate _cfg = \e -> do
-  run $ case e of
+handler bState _cfg e = run $ do
+  $(logTM) `debug` [int||Received Slack event: #{show @Text e}|]
+  case e of
     Command cmdType slashCmd -> case cmdType of
-      Fixtures.HelpCommand -> processHelpCommand slashCmd
-      unknownCmd           -> log' [int||Unknown command #{unknownCmd}|]
+      Fixtures.HelpCommand -> katipAddNamespaceText cmdType $ processHelpCommand slashCmd
+      unknownCmd           -> $(logTM) `warn` [int||Unknown command #{unknownCmd}|]
```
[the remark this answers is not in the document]
YuriRomanowski: If we use `logLocM`, we'll need to add `HasCallStack` to every function? If yes, I'd like to use template haskell.

src/TzBot/Logger.hs
```diff
@@ -0,0 +1,77 @@
...
+
+module TzBot.Logger
+  ( module TzBot.Logger
+  , module Katip
+  ) where
+
+import Universum
+
+import Data.Aeson (KeyValue((.=)), ToJSON(..), object)
+import Katip
```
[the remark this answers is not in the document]
YuriRomanowski: Yep, I decided so too :) Otherwise we get either too large import list or the module cluttered with one kind qualified imports.

src/TzBot/ProcessEvents.hs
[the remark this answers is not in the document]
Martoon-00: I think no, why? If a function has `HasCallStack` in its constraint, then its uses will get location of the call site.

src/TzBot/Util.hs
YuriRomanowski: Well, in this function we really would like to catch all exceptions (and asynchronous ones), is it bad? We can also add a haddock, e.g. "This function catches all the exceptions including asynchronous ones"

src/TzBot/ProcessEvents.hs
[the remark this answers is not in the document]
YuriRomanowski: Replaced logTM with logLocM, and it allowed me to make more convenient helpers, it looks awesome now :)

src/TzBot/Util.hs
YuriRomanowski: I guess the best thing we can do here is to throw away ExceptT IO and just use MonadThrow, but maybe it's better to implement in another PR

Martoon-00: > Well, in this function we really would like to catch all exceptions (and asynchronous ones),

If we try to abort the program then I suppose this thread should not catch the exception and report it as a problem, it is a normal situation.

> I guess the best thing we can do here is to throw away ExceptT IO and just use MonadThrow, but maybe it's better to implement in another PR

Agree, let's handle this separately.

src/TzBot/ProcessEvents.hs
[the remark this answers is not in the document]
Martoon-00: Eeee, really nice.

Martoon-00 (approved):
Great :+1:

PR #40, opened 2023-01-19
[#35] Fetch user info concurrently

### LINKED ISSUE #35: Fetch for user info concurrently

Currently, we run `getUser` sequentially on all the users of the channel, and that may cause large delays.

Probably we should get those concurrently.

Also, maybe asking user info and sending a notification could be placed in one thread, this way fetching information about one user won't affect other users. Although not sure if it is strictly necessary.

If you have other immediate fixes in the surrounding code, let's also apply them. However generally I would put care about performance to a separate ticket.

# Acceptance criteria

* We do not ask for user info sequentially in our handler for detected timestampts.

YuriRomanowski:
## Description
Problem: Currently, we run getUser sequentially on all the users of the channel, and that may cause large delays.

Solution: Put `getUserInfoCached` and `sendEphemeralMessage` in separate thread.

## Related issue(s)

Fixed #35 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/TzBot/ProcessEvent.hs
+++ b/src/TzBot/ProcessEvent.hs
@@ -94,14 +94,15 @@ processMessageEvent evt = when (newMessageOrEditedMessage evt) $ do
     let notBotAndSameTimeZone u = not (uIsBot u) && uTz u /= uTz sender
         notSender userId = userId /= uId sender
 
-    usersToNotify <- filter notBotAndSameTimeZone
-      <$> (mapM getUserCached $ filter notSender $ S.toList usersInChannelIds)
---    usersToNotify <- mapM getUserCached $ S.toList usersInChannelIds -- dev
-
-    forConcurrently_ usersToNotify $ \userToNotify -> do
-      let ephemeralMessage =
-            renderEphemeralMessageForOthers userToNotify ephemeralTemplate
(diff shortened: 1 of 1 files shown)


YuriRomanowski:
I wondered what will occur in the situation when multiple threads want to insert something into the cache, and tried to modify cache slightly such that threads didn't insert to the cache directly but inserted items into the `Chan`, which was read by a separate thread that is supposed to insert the read items sequentially, and and tried to write some benchmarks, but results seem to be almost the same. (my drafts for these benchmarks can be found in the branch `YuriRomanowski/compare-cache-modifications-draft`).

Martoon-00:
Hmhm, the library mentioned something about atomicity and about STM. It does not say explicitly it is thread-safe, but in the header it says to use STM under hood, so it should be good on itself.

OTOH, I think STM does not behave very well under high contention. Not like we expect it now, but in channels with 1000 members it may be the case.

I'd leave this as a separate PR, with the notes about benchmark results attached. Using two means for providing thread-safety is an overhead (not only in performance, but in code complexity too) that worries me, this is something that's worth discussing.

Martoon-00 (approved):

PR #42, opened 2023-01-23
[#38] Ignore time references in code blocks

### LINKED ISSUE #38: Ignore time references in codeblocks

Among our requirements for the timezone bot there is one that time references inside the code blocks should be ignored, otherwise it can make problems when someone shares a big log file. We should implement this feature, for this we can analyze the block structure of the incoming message that is provided by slack and ignore code blocks (see https://api.slack.com/events/message for reference).

# Acceptance criteria

* Time references inside code blocks are ignored.

YuriRomanowski:
## Description
Problem: The bot should ignore time references in code blocks because they are most probably some logs or just some code quotations.

Solution: Ignore codeblocks by analyzing (undocumented) block structure of incoming Slack messages, using some simple code ignoring function when the block structure cannot be parsed.

## Related issue(s)

Fixed #38

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -22,6 +22,7 @@ library:
   - universum
   - aeson
   - aeson-casing
+  - aeson-pretty
   - base >= 4.7 && < 5
   - bytestring
   - cache
@@ -29,7 +30,9 @@ library:
   - containers
   - directory
   - fmt
+  - deriving-aeson
   - dlist
+  - fmt
   - formatting
   - guid
   - http-client
@@ -56,6 +59,7 @@ library:
   - text
   - time
   - time-compat
+  - transformers
   - transformers-base
   - tz
   - tztime
@@ -88,20 +92,22 @@ tests:
     - -with-rtsopts=-N
     dependencies:
     - base >= 4.7 && < 5
+    - universum
     - tzbot
+    - aeson
     - containers
     - hspec
+    - http-types
+    - nyan-interpolation
+    - servant-client-core
+    - string-conversions
     - tasty
     - tasty-hspec
     - tasty-hunit
+    - tasty-quickcheck
     - time
     - tztime
     - QuickCheck
-    - http-types
-    - servant-client-core
-    - tasty-quickcheck
-    - time
-    - universum
   doctests:
     main:                doctests.hs
     source-dirs:         test
--- a/src/TzBot/ProcessEvents/Common.hs
+++ b/src/TzBot/ProcessEvents/Common.hs
@@ -4,18 +4,31 @@
 
 module TzBot.ProcessEvents.Common
   ( openModalCommon
+  , getTimeReferencesFromMessage
+
+    -- * exported for tests
+  , ignoreCodeBlocksManually
   ) where
 
 import Universum
 
+import Data.Aeson.Encode.Pretty (encodePrettyToTextBuilder)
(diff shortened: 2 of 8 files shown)


package.yaml
```diff
@@ -80,20 +83,22 @@ tests:
...
     - tasty-hunit
+    - tasty-quickcheck
     - time
     - tztime
     - QuickCheck
-    - http-types
-    - servant-client-core
-    - tasty-quickcheck
-    - time
-    - universum
```
Martoon-00, line 94: :+1:

src/TzBot/ProcessEvents/Common.hs
```diff
@@ -53,3 +63,70 @@ openModalCommon message channelId whoTriggeredId triggerId mkModalFunc = do
...
+["asd "," qwe ","` bad block"]
+>>> ignoreCodeBlocksManually "asd `little block` ```big code block1 ```"
+["asd "," "]
+>>> ignoreCodeBlocksManually "``` code ``` foo `bar` baz"
+[" foo "," baz"]
+
+(These are just examples, not tested by doctest)
+ -}
+ignoreCodeBlocksManually :: Text -> [Text]
+ignoreCodeBlocksManually txt =
```
Martoon-00, line 115: Probably I would also cover this function with some test.

Not like that would be handy now, but if it turns out that we need to make this function smarter, then tests would help to catch regressions.

Martoon-00 (approved):
Very cool, I very much like how the code is organized, the fact that you considered a backup manual method for ignoring blocks since the main method is not reliable, and that an elaborate test is present.

Leaving only one comment, but approving in advance, feel free to merge after resolving that comment.

src/TzBot/ProcessEvents/Common.hs
YuriRomanowski: Hm, there are some tests already, do you mean just to check them by doctest?

Martoon-00: Ah right, at that moment I forgot about the haddock.

Yes, is there a reason not to check it with doctest? (AFAIR it has been enabled, or I'm mistaken?)

YuriRomanowski: I didn't want to export this function, AFAIR this is required for `doctest` to work

Martoon-00: Yep, I see...

Still, I think it would be really nice to be tested. I'm afraid that the presence of doctests may add to (false) confidence in that the function keeps working properly - then someone needs to change it :roll_eyes: 

Usually, in such cases we either put the function into internal module, or make its name such scary that the person would think twice before using it. It already has "manually" word, which would even seem enough for me.

What do you think?

YuriRomanowski: > It already has "manually" word, which would even seem enough for me.

That's very funny note :smile: 

Hm, it seems you are right. This module is not a part of some library, so there is no strong need in `Internal` modules, so good naming and proper comment will be enough.

> Still, I think it would be really nice to be tested.

Sure, let's just replace these haddock examples with ordinary tests.

Martoon-00 (approved):
Nice :+1:

PR #45, opened 2023-01-26
[Fix] Fix time reference parser

YuriRomanowski:
## Description
Problem: The parser is not perfect and can be improved.

Solution: Try to improve the parser; add preceding tokenization, add case-insensitive days of week, months and timezones, add constraints, some other fixes.

## Related issue(s)

Fixed #

- Tests
        silently reappearing again.

- Documentation
    - Haddock


src/TzBot/Lexer.hs
```diff
@@ -0,0 +1,443 @@
...
+
+>>> parseTimeRefs "10am UTC+03:00"
+[TimeReference {trText = "10am UTC+03:00", trTimeOfDay = 10:00:00, trDateRef = Nothing, trLocationRef = Just (OffsetRef 180)}]
+
+>>> parseTimeRefs "10am UTC+3"
+[TimeReference {trText = "10am UTC+3", trTimeOfDay = 10:00:00, trDateRef = Nothing, trLocationRef = Just (OffsetRef 180)}]
+
+>>> parseTimeRefs "Let's meet between 10am and 11:30am"
+[TimeReference {trText = "10am", trTimeOfDay = 10:00:00, trDateRef = Nothing, trLocationRef = Nothing},TimeReference {trText = "11:30am", trTimeOfDay = 11:30:00, trDateRef = Nothing, trLocationRef = Nothing}]
+
```
YuriRomanowski, line 88: The new tests start here

src/TzBot/Lexer.hs
```diff
@@ -0,0 +1,443 @@
...
+  , ("July", 7)
+  , ("August", 8)
+  , ("September", 9)
+  , ("October", 10)
+  , ("November", 11)
+  , ("December", 12)
+  ]
+
+addCut3Option :: (Text, b) -> [(Text, b)]
+addCut3Option t = [t, first (T.take 3) t]
```
Martoon-00, line 359: Surprising that it just really works for months and days of week, never payed attention to that fact.

src/TzBot/Lexer.hs
```diff
@@ -0,0 +1,443 @@
...
+        spacedComma
+        optional' (word' "the" >> space)
+        dayOfMonth <- fst <$> dayOfMonthParser
+        pure $ DayOfMonthRef dayOfMonth monthOfYear
+    ]
+
+dayOfMonthParser :: TzParser (DayOfMonth, Bool)
+dayOfMonthParser = do
+  n <- number
+  guard (n < 32)
```
Martoon-00, line 332: Wet dreams: If we could send messages only to the author, in case this condition violates I would avoid parsing error and rather troll the author back :)

src/TzBot/Lexer.hs
```diff
@@ -0,0 +1,443 @@
...
+  , ("Wednesday", Wednesday)
+  , ("Thursday", Thursday)
+  , ("Friday", Friday)
+  , ("Saturday", Saturday)
+  , ("Sunday", Sunday)
+  ]
+
+dayOfMonthRefParser :: TzParser DateReference
+dayOfMonthRefParser = do
+  _ <- optional' (word' "on" >> space)
```
Martoon-00, line 310: I would actually add `at` and `in` too. That would be not correct English, but we should anticipate expectations.

src/TzBot/Lexer.hs
```diff
@@ -0,0 +1,443 @@
...
+  choice' $
+    [ do
+        _ <- optional' (word' "the" >> space)
+        (dayOfMonth, hasSuffix) <- dayOfMonthParser
+        _ <- optional' (space >> word' "of")
+        monthOfYear <- optional' (space >> monthOfYearParser)
+        -- otherwise we can consider a random number
```
Martoon-00, line 317: Better say "unrelated", not "random", otherwise this sentence can be misinterpreted

src/TzBot/Lexer.hs
```diff
@@ -0,0 +1,443 @@
...
+  isAm <- if isJust maybeMin
+          then fromMaybe True <$> optional' isAmParser
+          else isAmParser
+
+  let todSec = 0
+      todHour
+        | isAm = hour
+        -- pm here
+        | hour < 12 = hour + 12
+        -- ignore pm if hour > 12
```
Martoon-00, line 250: Such cases make me think that it would really be nice to make parsers report errors, not only let them accept or deny values.

However it's hard to imagine that people would often type `13pm`

tzbot.cabal
```diff
@@ -115,12 +116,14 @@ library
...
     , bytestring
     , cache
+    , case-insensitive
     , clock
     , containers
     , directory
     , dlist
     , fmt
     , formatting
+    , glider-nlp
```
Martoon-00, line 126: Looks like more dark magic leaked into here

Martoon-00 (commented):
Looks nice, this all is a decent improvement.

Looking forward to seeing it finished, if you are planning to add more things.

Leaving a few small comments on things that I spotted.

src/TzBot/Lexer.hs
YuriRomanowski: As a user, I would like the bot not to be annoying in any extent, rather to be helpful. There is no worth for user reporting his misprint, and also we can report false positives quite often. So I think we should report only those errors which can be helpful to be reported, like we do this with time shifts, and be humble for other errors. For the latter better reaction would be either no reaction at all, or an attempt to fix it; if it's unsuccessful, we lose nothing, because the initial input was already broken.

Martoon-00: > There is no worth for user reporting his misprint

In other cases I could argue, but for time... thinking about it, that sounds reasonable.

> also we can report false positives quite often

Yeah, that's true.

tzbot.cabal
YuriRomanowski: Why?

src/TzBot/Lexer.hs
YuriRomanowski: I also thought about totally throwing away all the "stop-words", but some problems then appear with restoring the original message. But this idea would make the parsing much easier.

YuriRomanowski: Hm, but we can attach the list of stop words to the previous meaningful token, and then detach from it.

src/TzBot/Lexer.hs
```diff
@@ -0,0 +1,443 @@
...
+anyWord :: TzParser Text
+anyWord = token' $ \case
+  Word w -> pure w
+  _ -> empty
+
+tzRefParser :: TzParser LocationReference
+tzRefParser = do
+  region <- anyWord
+  _c <- punct '/'
+  place <- anyWord
```
YuriRomanowski, line 380: That actually not true: "America/Argentina/Buenos_Aires" is not parsed this way.

tzbot.cabal
Martoon-00: It is the library for natural languages processing AFAIU, is it used here?

YuriRomanowski: Yes, it has "nlp" in its name, but in fact it's very simple and straightforward library, we even could right that code ourselves. But it just works.

Martoon-00: Ah okay, its usage looks really fair then.

src/TzBot/Parser.hs
```diff
@@ -6,21 +6,23 @@ module TzBot.Parser
...
+import Data.Text qualified as T
 import Data.Time (DayOfWeek(..))
 import Data.Time.Calendar.Compat (DayOfMonth, MonthOfYear)
 import Data.Time.LocalTime (TimeOfDay(..))
-import Data.Time.Zones.All (tzNameLabelMap)
-import Text.Megaparsec (Parsec, anySingle, choice, match, parseMaybe, takeRest, try)
-import Text.Megaparsec.Char (char, digitChar, space, space1, string')
-import Text.Megaparsec.Char.Lexer (decimal)
+import Data.Time.Zones.All (TZLabel, tzNameLabelMap)
+import Glider.NLP.Tokenizer
```
Martoon-00, line 18: Imports

Martoon-00 (approved):
LGTM, leaving only one comment :+1:

Issue #41, opened 2023-01-23
Differentiate between dev/prod modes in config


Martoon-00:
# Clarification and motivation

In the code we have a couple of comments that you can uncomment to enable dev mode that allows testing without involving many people. That also will make code easier to maintain (since no code will be commented by default).

I think we should just make this configurable, that's something that may be useful for others. Configurable via config or CLI, what suits better (but probably not both, unless really both can be useful).

# Acceptance criteria

* It is possible to enable dev mode for already compiled executable.
* What exactly dev mode changes is documented.
* If there are any run/testing instructions, they are updated to mention dev mode.

PR #46, opened 2023-01-26
[#44] Add help command

### LINKED ISSUE #44: Fix a bunch of issues

I posted a few issues in [this Notion document](https://www.notion.so/serokell/Testing-iteration-1-from-dbfd12f048824fd68057f664fead248f), let's resolve them under this ticket.

# Acceptance criteria

* All the mentioned issues are discussed and resolved.

YuriRomanowski:
## Description
Problem: Currently, each ephemeral with translations contain button which functionality doubles existing entrypoint. Because it can be annoying, we should remove it and instead give a hint to a user how to call tzbot entrypoints.

Solution: Remove buttons from ephemerals, add command /tzhelp, append the usage of that command randomly with rather low chance to ephemerals with translations.

Also done:
* Add dev mode using undocumented envvar;
* Handle messages with unknown subtype as ordinary messages.

## Related issue(s)

Fixed #44, #41

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/TzBot/Slack/Fixtures.hs
+++ b/src/TzBot/Slack/Fixtures.hs
@@ -15,26 +15,22 @@ module TzBot.Slack.Fixtures
   , reportInputElementActionId
   , reportInputBlockId
 
-  -- * Report button inside ephemerals
-  , MessageIdentifier (..)
-  , reportFromEphemeralActionIdPrefix
-  , reportFromEphemeralActionId
-  , getMessageIdFromReportEphemeral
-  , reportFromEphemeralBlockId
-
   -- * Modal identifiers
   , viewModal
   , reportModal
+
+  -- * Help message
+  , pattern HelpCommand
+  , helpMessage
+  , helpUsage
   ) where
 
 import Universum
 
-import Data.Aeson (FromJSON, ToJSON)
-import Data.Text qualified as T
 import GHC.TypeLits (symbolVal)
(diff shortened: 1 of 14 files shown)


src/TzBot/Slack/Fixtures.hs
```diff
@@ -95,3 +67,30 @@ viewModal = "view_modal"
...
+Whenever the message is posted in the channel I'm in, if this \
+message contains any time references, I will translate them and send it \
+to you via an ephemeral message. If a time reference is invalid for some \
+reason, I will say what's wrong with it.
+Also you can trigger the message translation yourself, \
+by using entrypoints in the Slack message context menu; this works \
+in every channel of the workspace (including direct messages). \
+There you also can report the bot working wrongly, \
+we will be very grateful for your feedback!
+|]
```
Martoon-00, line 96: You can use a special modifier to avoid translating line feeds from the code to the output: `[int|n|`. With it only blank lines will produce line feeds, this seems to be less error-prone for the possible future edits.

And `!` modifier to quickly check that the output will be what you expect.

Martoon-00 (approved):
This is a way to resolve that issue, pretty nice.

Also cool that you separated main change and the miscellaneous changes :+1:   But note that commit subject should not end with a dot. 

Leaving one comment.

PR #47, opened 2023-01-27
[#18] Check for errors resilience

### LINKED ISSUE #18: Make sure the app is very resilient to errors

After implementing the main logic, let's take extra measures to make sure that the application does not crash due to networking or internal errors.

# Acceptance criteria

* Some appropriate threading / errors handling model is introduced to make the app never crash.
* It should still be possible to send `SINGINT` and stop the app.

YuriRomanowski:
Problem: Currently, if unknown exception occurs (not BotException), the bot just goes down.

Solution: Add exception handlers, run graceful shutdown handler on UserInterrupt.

## Related issue(s)

Fixed #18 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/TzBot/BotMain.hs
+++ b/src/TzBot/BotMain.hs
@@ -14,7 +14,7 @@ import Network.HTTP.Client.TLS (tlsManagerSettings)
 import Options.Applicative (execParser)
 import Slacker
   (defaultSlackConfig, handleThreadExceptionSensibly, runSocketMode, setApiToken, setAppToken,
-  setOnException)
+  setGracefulShutdownHandler, setOnException)
 import System.Directory (doesFileExist)
 import Text.Interpolation.Nyan (int, rmode')
 
@@ -67,10 +67,15 @@ run opts = do
     let fifteenPercentAmplitudeSettings = defaultTzCacheSettings
           { tcsExpiryRandomAmplitudeFraction = Just 0.15
           }
+
+    gracefulShutdownContainer <- liftIO $ newIORef $ (pure () :: IO ())
+    let extractShutdownFunction :: IO () -> IO ()
+        extractShutdownFunction = writeIORef gracefulShutdownContainer
     let sCfg = defaultSlackConfig
               & setApiToken (unBotToken cBotToken)
               & setAppToken (unAppLevelToken cAppToken)
               & setOnException handleThreadExceptionSensibly -- auto-handle disconnects
+              & setGracefulShutdownHandler extractShutdownFunction
 
     bsManager <- liftIO $ newManager tlsManagerSettings
     bsMessagesReferences <- newIORef M.empty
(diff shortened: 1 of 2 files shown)


src/TzBot/BotMain.hs
```diff
@@ -67,10 +67,15 @@ run opts = do
     let fifteenPercentAmplitudeSettings = defaultTzCacheSettings
           { tcsExpiryRandomAmplitudeFraction = Just 0.15
           }
+
+    gracefulShutdownContainer <- liftIO $ newIORef $ (pure () :: IO ())
+    let extractShutdownFunction :: IO () -> IO ()
+        extractShutdownFunction = writeIORef gracefulShutdownContainer
```
YuriRomanowski, line 73: This looks _extremely_ strange, but the `gracefulShutdownHandler` field of the slacker config has type `IO () -> IO ()` and is called on slacker start, passed a `shutdownSocketMode` which actually performs shutdown. I have come up with anything better than just extract that function into the `IORef` :rofl: What am I doing wrong?

src/TzBot/ProcessEvents.hs
```diff
@@ -84,11 +85,17 @@ handler bState _cfg e = run $ do
...
+      eithRes <- catchAllErrors action
+      whenLeft eithRes $ \eithErr -> do
+        let logError :: (Exception e) => e -> BotM ()
+            logError err = $(logTM) `error_` [int||Error occured: #s{err}|]
+        case eithErr of
+          Left someExc
+            | Just UserInterrupt <- fromException someExc ->
+                liftIO $ join $ readIORef shutdownRef
+            | otherwise -> logError someExc
+          Right botExc -> logError botExc
```
YuriRomanowski, line 98: Also this `catchAllErrors` looks strange too, I had to add it because we can get an error unwrapping `BotM` action, but we still have to take other "flying around" unknown exceptions into account. But now I don't know how to handle this more elegantly.

Martoon-00: Looks tolerable for me for now :+1: 

I guess it will be improved in https://github.com/serokell/tzbot/issues/51.

Martoon-00 (approved):
Looks good overall, I don't see what can be added to this. Except for one thing.

Sad that the Slacker library handles exceptions/threading in its own contrived way as far as I can tell, and we get the following problems from that:
* Async exception will fall into handler, aborting the logic that is going on there; this could happen more gracefully, but ok for now.
* That juggling with shutdown handler (I think we really don't want to rely on OS signals handling, maybe only at top level of the code)
* I once managed to get `Ctrl + C` consumed by the app without any reaction.
* In fact, any accidentally uncaught exception will kill the internal socket machinery not gracefully. And the events loop is itself unprotected, so an exception thrown at idle time would result into dirty kill; but I may be mistaken here and I likely this is handled somehow.

Probably we could wrap `runSocketMode` into layer of code that would make it safer, I can suggest how we could improve that. However this does not seem to be crucial to fix, I think we can perfectly leave with what we have now.

Martoon-00 (approved):

PR #48, opened 2023-01-29
[#36] Express doubts in results

### LINKED ISSUE #36: Think about how to express doubts in the result

Imagine a scenario: today (at Jan 16) I write a message that mentions 12:00 of Nov 23. And we should try to guess an year.
We already have a guessing logic after #2, and the question is: how should we express when we are not sure about the result in such cases?

And there are options:
1. Never doubt, just post the guess.
2. Add a note that we are unsure.
3. Do not send any messages to other users, say the message author's that his timestamp is ambiguous and should be fixed.
4. Select between (2) and (3) depending on how sure we are in the result.

# Acceptance criteria

* We discussed the mentioned point and applied a solution.

YuriRomanowski:
## Description

Problem: In many cases we can't know the full date because we are not
able to get all the context of the time reference. In such situations we
should explain the user somehow that we are not 100% sure in the
inferred date.
    
Solution: Implement simple guessing logic; we are either totally sure or
sure only about time, output messages are changed correspondingly.

Also done: slightly changed messages sending according to [requirements](https://www.notion.so/serokell/Requirements-7dc47068a206498989c4b899a767d208): ephemeral is also sent to the sender himself if the message contains time references in another timezone than the sender's one.

## Related issue(s)

Fixed #36 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/TzBot/TimeReference.hs
+++ b/src/TzBot/TimeReference.hs
@@ -12,6 +12,7 @@ import Data.Time
   (DayOfWeek, LocalTime(localDay), NominalDiffTime, TimeOfDay, TimeZone, UTCTime, addLocalTime,
   diffLocalTime, localTimeToUTC, nominalDay, toGregorian, utc, utcToLocalTime)
 import Data.Time.Calendar.Compat (DayOfMonth, MonthOfYear)
+import Data.Time.Clock (diffUTCTime)
 import Data.Time.TZInfo qualified as TZI
 import Data.Time.TZTime qualified as TZT
 import Data.Time.Zones.All (TZLabel)
@@ -82,6 +83,23 @@ utcToUtcLocalTime = utcToLocalTime utc
 utcLocalTimeToUTC :: LocalTime -> UTCTime
 utcLocalTimeToUTC = localTimeToUTC utc
 
+convertUtcToOffsetTime :: Offset -> UTCTime -> LocalTime
+convertUtcToOffsetTime offset utcTime =
+  addLocalTime (offsetToNominalDiffTime offset) (utcToUtcLocalTime utcTime)
+
+convertOffsetTimeToUtc :: Offset -> LocalTime -> UTCTime
+convertOffsetTimeToUtc offset localTime =
+  utcLocalTimeToUTC $ addLocalTime (negate $ offsetToNominalDiffTime offset) localTime
+
+theSameDayInTwoTimezones :: UTCTime -> TZLabel -> Either TZLabel Offset -> Bool
+theSameDayInTwoTimezones timestamp senderTz eithTzOffset = do
+  let convertWithTzLabel tzL = TZT.tzTimeLocalTime $ TZT.fromUTC (TZI.fromLabel tzL) timestamp
+      localTime = ($ eithTzOffset) $ either
+        convertWithTzLabel
+        (\o -> convertUtcToOffsetTime o timestamp)
+      senderLocalTime = convertWithTzLabel senderTz
+  ((==) `on` localDay) senderLocalTime localTime
+
 {- | Converts a time reference to a moment in time (expressed in UTC).
 
   If the time reference contains a timezone abbreviation, and if that abbreviation\
@@ -103,29 +121,33 @@ timeReferenceToUTC sendersTZLabel eventTimestamp TimeReference {..} =
       -- In the case of rigid offset we don't need the `modifyLocal` from
       -- from the `tztime` package because there are no timeshifts that
       -- we should take into account. So we just use plain LocalTime.
-      let eventTimeUTC = utcToUtcLocalTime eventTimestamp
-          offsetNominal = offsetToNominalDiffTime offset
-      let refTimeUTC = eventTimeUTC & (
-            addLocalTime offsetNominal
+      let refTimeUTC = eventTimestamp & (
+            convertUtcToOffsetTime offset
             >>> dayTransition
             >>> TZT.atTimeOfDay trTimeOfDay
-            >>> addLocalTime (negate offsetNominal)
+            >>> convertOffsetTimeToUtc offset
             )
-      TRTUSuccess (utcLocalTimeToUTC refTimeUTC) (Right offset)
+      let dateInferred = isDateInferred eventTimestamp sendersTZLabel (Right offset) trDateRef refTimeUTC
+
+      TRTUSuccess $ TimeRefSuccess dateInferred refTimeUTC (Right offset)
     Right (Left (tzLabel, implicitSenderTimezone)) -> do
       let eventLocalTime = TZT.fromUTC (TZI.fromLabel tzLabel) eventTimestamp
       let eithRefTime = eventLocalTime & TZT.modifyLocalStrict (
-            dayTransition >>> TZT.atTimeOfDay trTimeOfDay
+            dayTransition
+            >>> TZT.atTimeOfDay trTimeOfDay
             )
       case eithRefTime of
         Left err -> tzErrorToResult implicitSenderTimezone tzLabel err
-        Right refTime -> TRTUSuccess (TZT.toUTC refTime) (Left tzLabel)
+        Right refTime -> do
+          let refTimeUTC = TZT.toUTC refTime
+          let dateInferred = isDateInferred eventTimestamp sendersTZLabel (Left tzLabel) trDateRef refTimeUTC
+          TRTUSuccess $ TimeRefSuccess dateInferred refTimeUTC (Left tzLabel)
 
   where
   tzErrorToResult :: Bool -> TZLabel -> TZT.TZError -> TimeReferenceToUTCResult
   tzErrorToResult implicitSenderTimezone tzLabel = \case
-    TZT.TZOverlap {} -> TRTUAmbiguous implicitSenderTimezone tzLabel
-    TZT.TZGap {}     -> TRTUInvalid implicitSenderTimezone tzLabel
+    TZT.TZOverlap {} -> TRTUAmbiguous $ TimeShiftErrorInfo implicitSenderTimezone tzLabel
+    TZT.TZGap     {} -> TRTUInvalid $ TimeShiftErrorInfo implicitSenderTimezone tzLabel
 
   -- This doesn't include setting time, only date changes
   dayTransition :: LocalTime -> LocalTime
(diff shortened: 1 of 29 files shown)


src/TzBot/TimeReference.hs
```diff
@@ -155,6 +177,31 @@ timeReferenceToUTC sendersTZLabel eventTimestamp TimeReference {..} =
...
+  -- "10am UTC tomorrow"?
+  -- So we should return True if it's the same day for both user's timezone and
+  -- for the mentioned timezone/offset.
+  Just (DaysFromToday _) -> not $ theSameDayInTwoTimezones eventTime tzLabel eithTzOffset
+  -- Here we always assume future day of week, but really sure if less than two days left
+  -- to that time.
+  Just (DayOfWeekRef _) -> (refTime `diffUTCTime` eventTime) > 2 * nominalDay
+  -- Here, if the month is known, we only don't know the year,
+  -- but most probably we'll guess right.
+  Just (DayOfMonthRef _ mbMonth) -> isNothing mbMonth
```
Martoon-00, line 203: Interesting :+1:

src/TzBot/TimeReference.hs
```diff
@@ -208,19 +255,29 @@ chooseBestYear dayOfMonth monthOfYear now = do
...
+  , trsEithTzOffset   :: Either TZLabel Offset
+    -- ^ The timezone or offset that this TimeReference is related to.
+    -- When the `TimeReference` does not explicitly mention a timezone/offset,
+    -- we assume it's related to the sender's timezone.
+  } deriving stock (Eq, Show)
+
+data TimeShiftErrorInfo = TimeShiftErrorInfo
+  { tseiIsImplicitSenderTimezone :: Bool
+  , tseiRefTimeZone :: TZLabel
+  } deriving stock (Eq, Show)
```
Martoon-00, line 272: :+1:

src/TzBot/TimeReference.hs
```diff
@@ -155,6 +177,31 @@ timeReferenceToUTC sendersTZLabel eventTimestamp TimeReference {..} =
...
+  -- This identifies the date clearly if the user uses only his own timezone.
+  -- But things get harder if the user mentions relative date ref and other
+  -- timezone (e.g. UTC). What date is meant if for example it's 2:00 now and
+  -- the user's timezone is UTC+3, so the UTC time is 23:00 and he writes
+  -- "10am UTC tomorrow"?
+  -- So we should return True if it's the same day for both user's timezone and
+  -- for the mentioned timezone/offset.
+  Just (DaysFromToday _) -> not $ theSameDayInTwoTimezones eventTime tzLabel eithTzOffset
+  -- Here we always assume future day of week, but really sure if less than two days left
+  -- to that time.
```
Martoon-00, line 199: The function name seems to do not what can be suggested from its name :thinking:

test/Test/TzBot/RenderSpec.hs
```diff
@@ -0,0 +1,170 @@
...
+  let [timeRef] = parseTimeRefs refText
+      ephemeralTemplate =
+        renderTemplate asForMessageM eventTimestamp sender $
+          NE.singleton timeRef
+
+      getBlocks senderFlag user = maybe [] (renderSlackBlocks senderFlag . Just) $ renderAllTP user ephemeralTemplate
+      senderBlocks = getBlocks asForSenderS sender
+      otherUserBlocks = getBlocks asForOthersS otherUser
+  senderBlocks @?= expectedSenderBlocks
+  otherUserBlocks @?= expectedOtherUserBlocks
```
Martoon-00, line 170: Very nice to have elaborate tests :+1:

test/Test/TzBot/RenderSpec.hs
```diff
@@ -0,0 +1,170 @@
...
+
+  , testCase "Close day of week" $ mkTestCase arbitraryTime1 "10am in Europe/Helsinki tuesday" userMoscow userHavana
+
+      (mkBlocks "\"10am in Europe/Helsinki tuesday\":" "*11:00 in Europe/Moscow, Tuesday, January 31*")
+      (mkBlocks "\"10am in Europe/Helsinki tuesday\":" "*03:00 in America/Havana, Tuesday, January 31*")
+
+  , testCase "Far day of week" $ mkTestCase arbitraryTime1 "10am in Europe/Helsinki thursday" userMoscow userHavana
+
+      (mkBlocks "\"10am in Europe/Helsinki thursday\":" "*11:00 in Europe/Moscow* (possibly, Thursday, February 02)")
+      (mkBlocks "\"10am in Europe/Helsinki thursday\":" "*03:00 in America/Havana* (possibly, Thursday, February 02)")
```
Martoon-00, line 70: Btw, if we put only the day next to "possibly" and not time, it may make an impression that the time is for sure correct. In the majority of cases it is, but sometimes time depends on the day.

However in the current way it's really convenient :thinking:   We could be smarter and see whether the alternative days result in different times or not, but that would probably be an overcomplication for our logic.

Martoon-00 (commented):
Looks good :+1: 

Only a couple of comments.

src/TzBot/TimeReference.hs
YuriRomanowski: Why do you think so?

test/Test/TzBot/RenderSpec.hs
YuriRomanowski: You are right, in this PR we really don't possible timeshifts in a comprehensive way, but it requires more effort for minor issue, and I tried to handle the most basic cases for now; how about creating a new issue for this?

Martoon-00: Yes, that sounds good.

Maybe for now we could print it as "11:00 in Europe/Moscow, Tuesday, February 02 (possibly)", or something similar? This way we do no unsafe claims, users can rely on the time if they notice the regularity and gonna take the risk.

YuriRomanowski: Do you mean that this way "possibly" covers the whole translation?

Martoon-00: Yes

YuriRomanowski: Well, if we don't take time shifts into account, the time itself should be correct, and in the other stuff we are really unsure, so I wanted to emphasize that we are unsure in *date*, not time.

YuriRomanowski: Also, "(possibly)" at the end of the text looks rather strange :smile:

Martoon-00: > Also, "(possibly)" at the end of the text looks rather strange smile

That's true, I'm not sure which form would be better to make "possibly" relate to the entire timestamp.

> if we don't take time shifts into account, the time itself should be correct

You handled time shifts in the core part very carefully, and now seem to neglect that effort by allowing invalid time because of date mismatch...

YuriRomanowski: > You handled time shifts in the core part very carefully, and now seem to neglect that effort by allowing invalid time because of date mismatch...

Well, we can do next: if we are not sure in date, we draw up a list of other candidates and check whether the time there is the same; if not, add a note (both to sender and receiver) that a time shift (possibly in both sender and receiver timezones) occurs, and suggest the sender, if the date was inferred wrong, to select the date explicitly. What do you think?

Martoon-00: That seems working, but probably would be relatively difficult? If not, let's really go for it.

YuriRomanowski: Well, things related to time shifts are always subtle :smile: But no problem, we'll just handle it.

test/Test/TzBot/RenderSpec.hs
```diff
@@ -0,0 +1,170 @@
...
+
+  , testCase "Inferred today" $ mkTestCase arbitraryTime1 "10am in Europe/Helsinki" userMoscow userHavana
+
+      (mkBlocks "\"10am in Europe/Helsinki\":" "*11:00 in Europe/Moscow* (possibly, Monday, January 30)")
+      (mkBlocks "\"10am in Europe/Helsinki\":" "*03:00 in America/Havana* (possibly, Monday, January 30)")
+
+  , testCase "Same timezone" $ mkTestCase arbitraryTime1 "10am in America/Havana" userMoscow userHavana
+
+      (mkBlocks "\"10am in America/Havana\":" "*18:00 in Europe/Moscow* (possibly, Monday, January 30)")
+      (mkBlocks "\"10am in America/Havana\":" "*your timezone is the same*")
```
dcastro, line 60: Should the user even be notified in this situation? I think it's ok not to notify them here.

dcastro: Let me elaborate:

Say user X is in timezone A, and user Y in timezone B.

* When user X says "10am in A": only user Y will be notified. User X will not receive any notification.
* When user X says "10am in B": following the same logic, I would argue _only_ the user X should be notified, and that Y should not receive any notification.

In other words, I think the logic should be consistent: _all_ users who are _not_ in the mentioned timezone get a notification, all users who are in the mentioned timezone do not get a notification. It doesn't matter whether the user is the sender or a receiver.

Martoon-00: Probably our last decision on this was to not be smart on omitting notifications, and maybe handle this separately later. Partially because there are corner cases here that we previously [discussed](https://github.com/serokell/tzbot/issues/4#issuecomment-1298473166), partially because in early development stages silent Bot can be considered as a misbehaving bot. 

Although this "later" has come, and there is no even a ticket created for this :sweat_smile:  Created #53 for handling this.

YuriRomanowski: > I think it's ok not to notify them here.

I wanted to post some thoughts, but @Martoon-00 was first, in fact outlining my opinion [here](https://github.com/serokell/tzbot/issues/53) :smile:

test/Test/TzBot/RenderSpec.hs
```diff
@@ -0,0 +1,170 @@
...
+
+userHavana :: User
+userHavana = User {uId="hav", uIsBot=False, uTz=America__Havana}
+
+test_renderSpec :: TestTree
+test_renderSpec = TestGroup "Render" $
+  [ testCase "Implicit sender's timezone" $ mkTestCase arbitraryTime1 "10am" userMoscow userHavana
+
+      []
+      (mkBlocks "\"10am\" in Europe/Moscow:" "*02:00 in America/Havana* (possibly, Monday, January 30)")
```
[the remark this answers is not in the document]
YuriRomanowski: Hm, that's very cool. How then we are supposed to express our doubts in this scheme? Is it done implicitly, just everything that is to the left but not in quotes is assumed? What explicit option would you suggest?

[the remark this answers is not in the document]
dcastro: > How then we are supposed to express our doubts in this scheme? 

Not sure what you mean by "doubts" here. There's explicit information (what the sender writes) and inferred information (i.e. guesses, like "the user didn't say which day he meant, I guess he meant 'today'").

> Is it done implicitly, just everything that is to the left but not in quotes is assumed? 

Yes, exactly. Everything that is to the left but not in quotes is information we inferred. Like when the user simply says "10am", we infer that he must have meant 10am in their current timezone.

PR #54, opened 2023-02-01
[#22] Revise messages editing

### LINKED ISSUE #22: Handle messages editing
"Message edited" event was already mentioned in an [issue #4](https://github.com/serokell/tzbot/issues/4). Let's discuss how we handle messages editing in more details here

Original suggestion was, given a message, to memorize what references we already processed from that link and process only new references that appeared upon editing and report these references in a new ephemeral message. What I don't quite like in this approach: our ephemeral messages needs a context of the original message, and so the latter should be somewhere close to the ephemeral one. And if some messages were added after the original one, and then it was edited, then the ephemeral is going to appear "in the open field", and it can be hard to figure out what it is related to. To solve this problem, we could for example add a reference to the original message on top of the ephemeral.

Also, instead, we could attempt to keep the corresponding ephemeral message just right after the original one, and edit it periodically. But the Slack docs [say](https://api.slack.com/messaging/managing#ephemeral) that it's not possible to edit an ephemeral directly, only on some user interaction with its interactive blocks (see [link](https://api.slack.com/messaging/modifying#updating_response_url)). So, one possible solution is:
* When someone posts a message with time references, we send ephemerals to all the users in the channel
* When an original message is edited, we memorize it's most fresh version in our bot;
* Ephemeral message that we posted has an interactive button, say, "Refresh", and when a user triggers it, if we have more recent version of that ephemeral, we replace old version with it;
* So, every user, if they see that the original messages was edited, can trigger "Refresh" button and get a fresh version of time references

The disadvantage of this approach is that if someone didn't understand that they should toggle the refresh button they could easily get wrong information from the ephemeral. But we can just explicitly write in the ephemeral "Toggle this button if the original message was edited" :smile: 

Another possible option that I thought about, is that, after getting a message with time reference, we can, instead of emitting any ephemerals immediately, just edit the original message, adding the interactive block to it, saying "translate to my timezone". After triggering this button, bot sends an ephemeral to this user with all needed information about time references. I think that this version is more convenient because the user controls all the flow himself. The only disadvantage is that we have to edit the original message, and sometimes it's important to know whether this message was edited by the sender or not.

What do you think guys? I didn't try how can these ideas implemented technically, but from docs it seems that we can do this. Let's discuss it here

# Acceptance criteria

* Discuss how messages editing should be handled
* Implement the best idea

YuriRomanowski:
Problem: Currently message_changed event is handled by translating new time references and sending ephemeral with them, which can be inconvenient sometimes.

Solution: On message_changed event send an ephemeral containing message permalink and all message time references translated.

## Related issue(s)

Fixed #22

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/TzBot/ProcessEvents/Message.hs
+++ b/src/TzBot/ProcessEvents/Message.hs
@@ -9,13 +9,14 @@ module TzBot.ProcessEvents.Message
 import Universum hiding (try)
 
 import Control.Concurrent.Async.Lifted (forConcurrently)
+import Data.List (singleton)
 import Data.List.NonEmpty qualified as NE
-import Data.Map qualified as M
 import Data.Set qualified as S
 import Data.Text.Lazy.Builder (Builder)
 import System.Random (randomRIO)
 import Text.Interpolation.Nyan (int, rmode', rmode's)
 
+import TzBot.Cache qualified as Cache
 import TzBot.Config (Config(..))
 import TzBot.Logger
 import TzBot.ProcessEvents.Common (getTimeReferencesFromMessage)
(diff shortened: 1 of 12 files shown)


src/TzBot/ProcessEvents/Message.hs
```diff
@@ -105,97 +77,163 @@ withSenderNotBot evt = do
...
+
+  handleMessageChanged :: BotM ()
+  handleMessageChanged = katipAddNamespaceText "edit" do
+    messageRefsCache <- asks bsMessageCache
+    mbMessageRefs <- Cache.lookup msgId messageRefsCache
+    -- if not found or expired, just ignore this message
+    -- it's too old or just didn't contain any time refs
+    whenJust mbMessageRefs $ \(_oldMsg, oldRefs) -> do
+      let originalRefsFound = not $ all (`elem` oldRefs) timeRefs
+      -- no original references found, ignoring
```
dcastro, line 160: I think here it would be clearer to name the var `newRefsFound` and change the comment to "no new references found". 

"Original refs" seems to suggest refs from the original message, as in, the old message before it was changed.

src/TzBot/ProcessEvents/Message.hs
```diff
@@ -105,97 +77,163 @@ withSenderNotBot evt = do
...
+  notSameTimeZone u = uTz u /= uTz sender
+  notBotAndSameTimeZone u = notBot u && notSameTimeZone u
+
+  getWhetherToShowHelpCmd :: BotM Bool
+  getWhetherToShowHelpCmd = do
+    inverseChance <- asks $ cInverseHelpUsageChance . bsConfig
+    liftIO $ fmap (== 1) $ randomRIO (1, inverseChance)
+
+  logNoTimeRefsFound = logInfo "No time references for processing found"
+  withNonEmptyTimeRefs trs action =
```
dcastro, line 120: Can you please add a type sig to `withNonEmptyTimeRefs`?

In general, I prefer having type sigs even on local functions, it helps with readability.

src/TzBot/RunMonad.hs
```diff
@@ -33,12 +32,11 @@ data BotState = BotState
   { bsConfig         :: BotConfig
   , bsManager        :: Manager
   , bsFeedbackConfig :: FeedbackConfig
-  -- TODO: after #22 bsMessagesReferences should either disappear or become
-  -- cached (not IORef).
-  , bsMessagesReferences :: IORef (M.Map MessageId (S.Set TimeReferenceText))
   , bsUserInfoCache  :: TzCache UserId User
   , bsConversationMembersCache :: TzCache ChannelId (S.Set UserId)
   , bsReportEntries  :: TzCache ReportDialogId ReportDialogEntry
+  , bsMessageCache   :: TzCache MessageId (Message, [TimeReference])
```
dcastro, line 38: Is the `Message` in the cache ever needed? It seems we only use `bsMessageCache` in `handleMessageChanged`, and there we only need the `[TimeReference]` and ignore the `Message`.

src/TzBot/ProcessEvents/Message.hs
```diff
@@ -105,97 +77,163 @@ withSenderNotBot evt = do
...
-        logInfo [int||#{errs} ephemerals failed|]
+        ephemeralsMailing channelId sendActionLocal
+
+  handleNewMessage :: BotM ()
+  handleNewMessage = do
+    withNonEmptyTimeRefs timeRefs $ \neTimeRefs -> do
+      -- save message only if time references are present
+      asks bsMessageCache >>= Cache.insert msgId (msg, timeRefs)
+      let ephemeralTemplate = renderTemplate now sender neTimeRefs
+      whenJust (renderErrorsForSenderTP ephemeralTemplate) $ \errorsMsg -> do
```
dcastro, line 181: Shouldn't we also call `renderErrorsForSenderTP` in `handleMessageChanged`?

src/TzBot/Slack.hs
```diff
@@ -156,6 +157,15 @@ updateModal req = do
...
 
+getMessagePermalinkCached :: ChannelId -> MessageId -> BotM Text
+getMessagePermalinkCached channelId msgId =
+  asks bsMessageLinkCache >>=
+    (Cache.fetchWithCacheRandomized msgId \msgId' -> do
+      token <- getBotToken
+      resp <- getPermalink token channelId msgId'
+      handleSlackErrorSingle "chat.getPermalink" resp)
+
+--
```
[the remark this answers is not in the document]
YuriRomanowski: Sometimes I want to add this as a delimiter for some reason :) But let's delete it here

src/TzBot/ProcessEvents/Message.hs
YuriRomanowski: Yes, you are right

src/TzBot/Slack.hs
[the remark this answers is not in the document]
dcastro: Oh, if you meant to create a "section", to organize declarations in groups, then I would advise doing [something like this](https://github.com/serokell/tztime/blob/60d531be17a6fd99a9a73f80b6682f5e371d5a04/src/Data/Time/TZTime.hs#L142-L144):

```hs
----------------------------------------------------------------------------
-- < section name > 
----------------------------------------------------------------------------

f :: ...

g :: ...

----------------------------------------------------------------------------
-- < section name > 
----------------------------------------------------------------------------

h :: ...
```

This makes the section much more visible and allows you to name the section (which makes the logic behind the grouping clear to the reader):

We use that pattern very often in some serokell projects, I even added it as a VisualCode snippet:

```
	"section": {
		"prefix": "section",
		"body": [
			"----------------------------------------------------------------------------",
			"-- $0",
			"----------------------------------------------------------------------------"
		]
	},
```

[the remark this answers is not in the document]
dcastro: I see there are a few other comments like `----` in this file (and perhaps other files?), so using those "sections" here would be a good idea.

dcastro (approved):
LGTM! 

My only "unresolved" comment is about sections, feel free to either apply them here or in another PR, I'm approving either way ^^

PR #55, opened 2023-02-01
[#51] Revise exceptions handling

### LINKED ISSUE #51: Revise exceptions handling

Currently we use `ExceptT BotException IO`, which is not really convenient because we need to handle `BotException` and other possible exceptions from IO separately; this is also considered by an anti-pattern by some folks (https://www.fpcomplete.com/blog/2016/11/exceptions-best-practices-haskell/). I think we should replace this by something else.

# Acceptance criteria

* Discuss how to change our way of exception handling
* Implement the solution

YuriRomanowski:
## Description
Problem: Currently we use `ExceptT BotException IO` which is inconvenient because both BotException, other sync exceptions and async exceptions should be handled separately.

Solution: Remove ExceptT, use UnliftIO for exceptions handling, also use UnliftIO for asyncs instead of monad-control.

## Related issue(s)

Fixed #51

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -43,12 +43,8 @@ library:
   - katip
   - lens
   - lens-aeson
-  - lifted-async
-  - lifted-base
   - managed
   - megaparsec
-  - monad-control
-  - mtl
   - nyan-interpolation
   - o-clock
   - random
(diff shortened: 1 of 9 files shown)


dcastro (approved):
LGTM! :)

PR #56, opened 2023-02-02
[Fix] Slight refactor of render part

YuriRomanowski:
## Description
We are going to work hard on translate/render modules and I would like some refactor to be present in all of that PRs beforehand.

Changes:
1. Remove button from section builders as it's never needed;
2. Add separate datatype for sender flag instead of plain Bool;
3. Replace plain text with markdown in the Slack section block;
4. Add functions for converting to offset time/back;
5. Unite fields of TimeReferenceToUtcResult for convenience.

## Related issue(s)

Fixed #

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/TzBot/Feedback/Save.hs
+++ b/src/TzBot/Feedback/Save.hs
@@ -19,7 +19,7 @@ import Data.Time.Zones.All (toTZName)
 import Text.Interpolation.Nyan (int, rmode')
 
 import TzBot.Logger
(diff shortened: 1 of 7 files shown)


Martoon-00 (approved):
I guess I have reviewed these changes already as part of separate PR.

Reminder to try to do each semantically different change in a separate commit, I guess doing so would also help to extract these refactoring changes out of the PR.

PR #57, opened 2023-02-02
[#36] Append inferred date parts

### LINKED ISSUE #36: Think about how to express doubts in the result

YuriRomanowski:
## Description

Problem: Full date is not shown for the original time reference, which
can be useful for receivers.
    
Solution: Append inferred parts of the date to the quoted original time
reference text.

## Related issue(s)

Fixed #36 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/TzBot/TimeReference.hs
+++ b/src/TzBot/TimeReference.hs
@@ -9,7 +9,7 @@ import Universum
 import Control.Arrow ((>>>))
 import Data.List.NonEmpty qualified as NE
 import Data.Time
-  (DayOfWeek, LocalTime(localDay), NominalDiffTime, TimeOfDay, TimeZone, UTCTime, addLocalTime,
+  (Day, DayOfWeek, LocalTime(localDay), NominalDiffTime, TimeOfDay, TimeZone, UTCTime, addLocalTime,
   diffLocalTime, localTimeToUTC, nominalDay, toGregorian, utc, utcToLocalTime)
 import Data.Time.Calendar.Compat (DayOfMonth, MonthOfYear)
 import Data.Time.TZInfo qualified as TZI
@@ -113,13 +113,15 @@ timeReferenceToUTC sendersTZLabel eventTimestamp TimeReference {..} =
       -- In the case of rigid offset we don't need the `modifyLocal` from
       -- from the `tztime` package because there are no timeshifts that
       -- we should take into account. So we just use plain LocalTime.
-      let refTimeUTC = eventTimestamp & (
+      let refTime = eventTimestamp & (
             convertUtcToOffsetTime offset
             >>> dayTransition
             >>> TZT.atTimeOfDay trTimeOfDay
-            >>> convertOffsetTimeToUtc offset
             )
-      TRTUSuccess $ TimeRefSuccess refTimeUTC (Right offset)
+      TRTUSuccess $ TimeRefSuccess
+        (convertOffsetTimeToUtc offset refTime)
+        (Right offset)
+        (localDay refTime)
     Right (Left (tzLabel, implicitSenderTimezone)) -> do
       let eventLocalTime = TZT.fromUTC (TZI.fromLabel tzLabel) eventTimestamp
       let eithRefTime = eventLocalTime & TZT.modifyLocalStrict (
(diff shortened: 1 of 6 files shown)


test/Test/TzBot/RenderSpec.hs
```diff
@@ -0,0 +1,71 @@
...
+        "03:00, Monday, 30 January 2023 in America/Havana"
+    ]
+
+  , testCase "Everything explicit" $
+    mkTestCase arbitraryTime1 "10am in America/Havana 3 Feb" userMoscow userHavana
+    [ translWithoutNotes
+        "\"10am in America/Havana 3 Feb\""
+        "10:00, Friday, 03 February 2023 in America/Havana"
+    ]
+  ]
```
dcastro, line 56: Please also add tests for these scenarios:

* Implicit date & explicit weekday: "10am on wednesday"
* Implicit date & explicit days from today: "10am in 3 days"
* Explicit day & implicit month: "10am on the 21st"

dcastro (approved):

src/TzBot/TimeReference.hs
```diff
@@ -113,29 +113,32 @@ timeReferenceToUTC sendersTZLabel eventTimestamp TimeReference {..} =
...
     Right (Left (tzLabel, implicitSenderTimezone)) -> do
-      let eventLocalTime = TZT.fromUTC (TZI.fromLabel tzLabel) eventTimestamp
-      let eithRefTime = eventLocalTime & TZT.modifyLocalStrict (
-            dayTransition >>> TZT.atTimeOfDay trTimeOfDay
-            )
+      let tzInfo = TZI.fromLabel tzLabel
+          eventTZTime = TZT.fromUTC (TZI.fromLabel tzLabel) eventTimestamp
+          refDate = dayTransition (TZT.tzTimeLocalTime eventTZTime)
+          eithRefTime = TZT.fromLocalTimeStrict tzInfo $
+            TZT.atTimeOfDay trTimeOfDay refDate
```
dcastro, line 128: I see you replaced `modifyLocalStrict` with a couple of back-and-forth conversions (`tzTimeLocalTime` and then `fromLocalTimeStrict`).

I'm guessing you did that so you could grab the invalid date in line 131:

```hs
        Left err ->
          tzErrorToResult implicitSenderTimezone tzLabel (localDay refDate) err
```

However, the invalid date is already contained inside the `err` variable, so i think it's more appropriate to fetch it from `err` instead, and continue using `modifyLocalStrict` as before:

```diff
-  tzErrorToResult :: Bool -> TZLabel -> Day -> TZT.TZError -> TimeReferenceToUTCResult
-  tzErrorToResult implicitSenderTimezone tzLabel refDay = \case
-    TZT.TZOverlap {} -> TRTUAmbiguous $
-      TimeShiftErrorInfo implicitSenderTimezone tzLabel refDay
-    TZT.TZGap {}     -> TRTUInvalid $
-      TimeShiftErrorInfo implicitSenderTimezone tzLabel refDay
+  tzErrorToResult :: Bool -> TZLabel -> TZT.TZError -> TimeReferenceToUTCResult
+  tzErrorToResult implicitSenderTimezone tzLabel = \case
+    TZT.TZOverlap invalidLocalTime _ _ -> TRTUAmbiguous $
+      TimeShiftErrorInfo implicitSenderTimezone tzLabel (localDay invalidLocalTime)
+    TZT.TZGap invalidLocalTime _ _ -> TRTUInvalid $
+      TimeShiftErrorInfo implicitSenderTimezone tzLabel (localDay invalidLocalTime)
```

dcastro (changes requested):

src/TzBot/TimeReference.hs
```diff
@@ -113,29 +113,32 @@ timeReferenceToUTC sendersTZLabel eventTimestamp TimeReference {..} =
...
             )
-      TRTUSuccess $ TimeRefSuccess refTimeUTC (Right offset)
+      TRTUSuccess $ TimeRefSuccess refTimeUTC (Right offset) (localDay refDate)
     Right (Left (tzLabel, implicitSenderTimezone)) -> do
-      let eventLocalTime = TZT.fromUTC (TZI.fromLabel tzLabel) eventTimestamp
-      let eithRefTime = eventLocalTime & TZT.modifyLocalStrict (
-            dayTransition >>> TZT.atTimeOfDay trTimeOfDay
-            )
+      let tzInfo = TZI.fromLabel tzLabel
+          eventTZTime = TZT.fromUTC (TZI.fromLabel tzLabel) eventTimestamp
```
dcastro, line 125: The expression ` TZI.fromLabel tzLabel` is repeated twice in lines 124 and 125.

Not a big deal though, after applying the refactoring I mentioned in my other comment, the variable `tzInfo` will no longer be needed anyway.

src/TzBot/TimeReference.hs
YuriRomanowski: Yes, you are right :+1:

dcastro (approved):
LGTM! ;)

PR #58, opened 2023-02-04
[#36] Warn user on possible timeshifts

YuriRomanowski:
## Description
Problem: In some cases the bot selects the date rather arbitrarily, and if a timeshift is somewhere around, the translation may be dependent on concrete date. We should warn a user about possible timeshifts.

Solution: For some types of date references, also compute a date range to check for the possible timeshift, and report to user if something found.

## Related issue(s)

Fixed part of #36

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/TzBot/TZ.hs
+++ b/src/TzBot/TZ.hs
@@ -1,0 +1,116 @@
+-- SPDX-FileCopyrightText: 2022 Serokell <https://serokell.io/>
+--
+-- SPDX-License-Identifier: MPL-2.0
+
+module TzBot.TZ
+  ( TimeShift (..)
+  , checkForTimeshifts
+  , checkForTimeshifts'
+  ) where
+
+import Universum
+
+import Data.Bits (shiftR)
+import Data.List qualified
+import Data.Time (TimeZone(TimeZone, timeZoneMinutes), UTCTime, addUTCTime, nominalDay)
+import Data.Time.Clock.POSIX (posixSecondsToUTCTime)
+import Data.Time.TZInfo (TZInfo(..), TZLabel(..))
+import Data.Time.TZInfo qualified as TZI
+import Data.Time.Zones (TZ)
+import Data.Time.Zones.Internal (utcTimeToInt64)
+import Data.Time.Zones.Types (TZ(..))
+import Data.Vector qualified as VB
+import Data.Vector.Unboxed qualified as VU
+
+import Data.List (nub)
+import TzBot.TimeReference (DateReference(..), TimeRefSuccess(..), TimeReference(..))
+import TzBot.Util (NamedOffset, Offset(..))
+
+-- | Represents a specific change in offset.
+data TimeShift = TimeShift
+  { tsShiftUtc :: UTCTime
+  -- ^ The time at which the offset change occurred / will occur.
+  , tsBefore   :: Offset
+  -- ^ The offset before the change.
+  , tsAfter    :: Offset
+  -- ^ The offset after the change.
+  , tsTzInfo   :: TZI.TZInfo
+  -- ^ The rules for the time zone where this change occurs.
+  } deriving stock (Eq, Show)
+
+-- | See the "Time shift warnings" section in `docs/development.md`.
+-- Sometimes the user does not specify a date, and we have to infer which date he meant.
+--
+-- When we do this, we should check if there were any time shifts around the inferred date,
+-- in either the "source" time zone or the receiver's time zone,
+-- and, if so, emit a warning.
+checkForTimeshifts :: TimeReference -> TimeRefSuccess -> TZLabel -> [TimeShift]
+checkForTimeshifts TimeReference{trDateRef} TimeRefSuccess{trsUtcResult, trsTzInfo} receiverTzLabel =
+  case trDateRef of
+    -- The date was inferred: we should check if there were any time shifts 3 days before/after the inferred date.
+    Nothing ->
+      checkEachTimeZone
+        (trsUtcResult & addUTCTime (nominalDay * -3))
+        (trsUtcResult & addUTCTime (nominalDay * 3))
+    -- The user specified the day of week; we had to infer which week they meant.
+    --
+    -- It's possible they meant "last week" instead, so we should check if
+    -- there were any time shifts between last week and the date we inferred.
+    Just (DayOfWeekRef {}) ->
+      checkEachTimeZone
+        (trsUtcResult & addUTCTime (nominalDay * -7))
+        trsUtcResult
+    -- The date was specified by the user; no need to perform the "time shift check"
+    Just (DaysFromToday {}) -> []
+    Just (DayOfMonthRef {}) -> []
+  where
+    -- Check for time shifts in both the "source" time zone and the receiver's time zone.
+    checkEachTimeZone :: UTCTime -> UTCTime -> [TimeShift]
+    checkEachTimeZone lowerBound upperBound =
+      nub [trsTzInfo, TZI.fromLabel receiverTzLabel]
+        & concatMap \tz -> checkForTimeshifts' tz lowerBound upperBound
+
+checkForTimeshifts' :: TZInfo -> UTCTime -> UTCTime -> [TimeShift]
+checkForTimeshifts' tzInfo (utcTimeToInt64 -> before) (utcTimeToInt64 -> after)
+  | before > after = []
+  | otherwise = do
+  let tz@(TZ trans _ _) = tziRules tzInfo
+      ixb = binarySearch trans before
+      startingFromIxb = VU.unsafeDrop ixb trans
+      desiredTimeshifts = zip [ixb..] $
+        takeWhile (<= after) $ VU.toList startingFromIxb
+      timeZones =
+        flip map desiredTimeshifts (\(ix, int64utc) ->
+          ( int64ToUtcTime int64utc
+          , Offset $ timeZoneMinutes $ timeZoneForIx tz ix)
+          )
+      timeShifts = zip timeZones $ Data.List.tail timeZones
+  map (\((_ub, b), (ua, a)) -> TimeShift ua b a tzInfo) timeShifts
+
+int64ToUtcTime :: Int64 -> UTCTime
+int64ToUtcTime = posixSecondsToUTCTime . fromIntegral
+{-# INLINE int64ToUtcTime #-}
+
+-- | Copied from "Data.Time.Zones" module of `tz` package because not exported.
+timeZoneForIx :: TZ -> Int -> NamedOffset
+timeZoneForIx (TZ _ diffs infos) i = TimeZone diffMins isDst name
+  where
+    diffMins = VU.unsafeIndex diffs i `div` 60
+    (isDst, name) = VB.unsafeIndex infos i
+{-# INLINE timeZoneForIx #-}
+
+-- | Copied from "Data.Time.Zones" module of `tz` package because not exported.
+-- Returns the largest index `i` such that `v ! i <= t`.
+--
+-- Assumes that `v` is sorted, has at least one element and `v ! 0 <= t`.
+binarySearch :: (VU.Unbox a, Ord a) => VU.Vector a -> a -> Int
+binarySearch v t | n == 1    = 0
+                 | otherwise = t `seq` go 1 n
+  where
+    n = VU.length v
+    go !l !u | l >= u = (l - 1)
+             | VU.unsafeIndex v k <= t  = go (k+1) u
+             | otherwise  = go l k
+      where
+        k = (l + u) `shiftR` 1
+{-# INLINE binarySearch #-}
--- a/src/TzBot/TimeReference.hs
+++ b/src/TzBot/TimeReference.hs
@@ -9,17 +9,13 @@ import Universum
 import Control.Arrow ((>>>))
 import Data.List.NonEmpty qualified as NE
 import Data.Time
-import Data.Time.Calendar.Compat (DayOfMonth, MonthOfYear, Year)
+import Data.Time.Calendar.Compat
 import Data.Time.TZInfo qualified as TZI
 import Data.Time.TZTime qualified as TZT
 import Data.Time.Zones.All (TZLabel)
-import Data.Time.Zones.Types (TZ(..))
-import Data.Vector qualified as V
-import Data.Vector.Unboxed qualified as VU
 import Formatting (Buildable)
-import Formatting.Buildable (Buildable(..))
-import Text.Interpolation.Nyan (int, rmode')
-import Text.Printf (printf)
+
+import TzBot.Util (Offset(..), tzInfoFromOffset)
 
 {- | An offset from UTC (e.g. @UTC+01:00@) with an optional timezone abbreviation (e.g. @BST@).
 
@@ -35,8 +31,6 @@ We use this type alias to make this distinction a bit more clear.
 
 (In fact, the @time@ package does not support timezones at all.)
 -}
-type NamedOffset = TimeZone
-
 type TimeReferenceText = Text
 
 -- | A reference to a point in time, e.g. "tuesday at 10am", "3pm CST on July 7th"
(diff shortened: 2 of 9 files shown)


src/TzBot/TZ.hs
```diff
@@ -0,0 +1,72 @@
...
+      desiredTimeshifts =
+        takeWhile (\(_ix, t) -> t <= after) $
+          zip [ixb..] $ VU.toList startingFromIxb
+      timeZones = fromJust $ nonEmpty $
+        flip map desiredTimeshifts (\(ix, int64utc) ->
+          ( int64ToUtcTime int64utc
+          , Offset $ timeZoneMinutes $ timeZoneForIx tz ix)
+          )
+      timeShifts = zip (toList timeZones) $ tail timeZones
+  map (\((_ub, b), (ua, a)) -> TimeShift ua b a) timeShifts
```
YuriRomanowski, line 45: If this function suites our need, we can then create an issue [in the original repo](https://github.com/ysangkok/haskell-tz/issues?q=is%3Aissue+is%3Aclosed) to add it there.

test/Test/TzBot/RenderSpec.hs
```diff
@@ -71,11 +74,38 @@ test_renderSpec = TestGroup "Render" $
...
         "02:00, Saturday, 21 January 2023 in America/Havana"
     ]
+
+  , TestGroup "TimeshiftWarning"
+    [ testCase "Implicit date" $
+      mkTestCase nearTimeshift "10am" userMoscow userHavana
+      [ translWithCommonNote
+          "\"10am\", 11 March 2023 in Europe/Moscow"
+          "02:00, Saturday, 11 March 2023 in America/Havana"
+          "_Warning! There is a time shift in America/Havana at 2023-03-12 05:00:00 UTC, so around this time conversion may depend on concrete date._"
```
dcastro, line 84: > There is a time shift in America/Havana at 2023-03-12 05:00:00 UTC

This doesn't seem to be correct. According to [this](https://www.timeanddate.com/time/zone/cuba/havana), the clocks are turned forward 1 hour from midnight to 1am.

dcastro:
I think I missed the conversation that led to this PR. Can you please point me to the context? 

> In some cases the bot selects the date rather arbitrarily, and if a timeshift is somewhere around, the translation may be dependent on concrete date. We should warn a user about possible timeshifts.

Also, you mentioned "in some cases", can you please give me a concrete example to see when/why this is useful? And how would you define "somewhere around"?

Sorry, I just think I'm missing some context to understand what this does and why :sweat_smile:

test/Test/TzBot/RenderSpec.hs
YuriRomanowski: Yes, I wrote "UTC", which means +5 hours relative to Havana. Did you take this into account in your comment?

YuriRomanowski: Actually, I must admit that using UTC for this maybe not so convenient for the end user, but it's the reliable thing for defining timeshift time. But I think we could just convert the timeshift UTCTime to the receiver's timezone, and show it instead. Does it make sense?

test/Test/TzBot/RenderSpec.hs
```diff
@@ -0,0 +1,119 @@
...
+          "03:00, Sunday, 12 March 2023 in America/Havana"
+          "_Warning! There is a time shift in America/Havana at 2023-03-12 05:00:00 UTC, so around this time conversion may depend on concrete date._"
+      ]
+    , testCase "Day of week, inversed users" $
+      mkTestCase nearTimeshift "10am on sunday" userHavana userMoscow
+      [ translWithCommonNote
+          "\"10am on sunday\", 12 March 2023 in America/Havana"
+          "17:00, Sunday, 12 March 2023 in Europe/Moscow"
+          "_Warning! There is a time shift in America/Havana at 2023-03-12 05:00:00 UTC, so around this time conversion may depend on concrete date._"
+      ]
```
dcastro, line 99: We should also add a test for when there is a time change in _both_ the source and the target time zones, e.g.:

```
    , testCase "Demo" $
      mkTestCase 
        (toUTC [tz|2023-03-25 15:00:00 [Europe/London]|])
        "10am"
        (User {uId="london", uIsBot=False, uTz=Europe__London})
        (User {uId="lisbon", uIsBot=False, uTz=Europe__Lisbon})
      [ translWithCommonNote
          "\"10am\", 26 March 2023 in Europe/London"
          "10:00, Sunday, 26 March 2023 in Europe/Lisbon"
          [int|s|
            _Warning! There is a time shift in Europe/London at 2023-03-26 01:00:00 UTC, so around this time conversion may depend on concrete date._
            _Warning! There is a time shift in Europe/Lisbon at 2023-03-26 01:00:00 UTC, so around this time conversion may depend on concrete date._
          |]
      ]
```

test/Test/TzBot/RenderSpec.hs
```diff
@@ -0,0 +1,119 @@
...
+        "02:00, Saturday, 21 January 2023 in America/Havana"
+    ]
+
+  , TestGroup "TimeshiftWarning"
+    [ testCase "Implicit date" $
+      mkTestCase nearTimeshift "10am" userMoscow userHavana
+      [ translWithCommonNote
+          "\"10am\", 11 March 2023 in Europe/Moscow"
+          "02:00, Saturday, 11 March 2023 in America/Havana"
+          "_Warning! There is a time shift in America/Havana at 2023-03-12 05:00:00 UTC, so around this time conversion may depend on concrete date._"
```
dcastro, line 84: From a user's point of view, it may not be clear what this means, what the implications are, or what the user should do about it:

> so around this time conversion may depend on concrete date

I think we should be very explicit about what the warning means and why we're warning the user about this time change. What do you think of this more verbose explanation?

> Warning: We inferred "10am" refers to 11 March 2023 in Europe/Moscow and converted it to America/Havana,
> but there is a time change near this date:
>
>   * at 01:00, 12 March 2023 in Europe/Moscow, the clocks are turned forward 1 hour(s).
>
> Beware that if this inferrence is not correct and the sender meant a different date, the conversion may not be accurate.


Now we're being very clear that we're emitting a warning _because_ we inferred the date, and that, in this specific case, the conversion is sensitive to this inference - if we inferred the wrong date, the conversion is likely to be wrong.

Note that we'd use 2 bullet points if there is a time change in _both_ the source and the target time zones, for example (I made up data for this example):

> Warning: We inferred "10am" refers to 11 March 2023 in Europe/Moscow and converted it to America/Havana,
> but there is a time change near this date:
>
>   * at 01:00, 12 March 2023 in Europe/Moscow, the clocks are turned forward 1 hour(s).
>   * at 02:00, 12 March 2023 in America/Havana, the clocks are turned forward 2 hour(s).
>
> Beware that if this inferrence is not correct and the sender meant a different date, the conversion may not be accurate.

test/Test/TzBot/RenderSpec.hs
dcastro: Let's also add tests that check that when the sender explicitly says "tomorrow", or "in 3 days", or explicitly specifies a date like "26 March", then the note is **not** emitted.

test/Test/TzBot/RenderSpec.hs
dcastro: Oh, I missed the "UTC" bit at the end :sweat_smile: 

> Actually, I must admit that using UTC for this is maybe not so convenient for the end user, but it's the reliable thing for defining timeshift time. But I think we could just convert the timeshift UTCTime to the receiver's timezone, and show it instead. Does it make sense?

I agree using UTC here is really inconvenient for users.

I think we should show the time in the timezone where the shift occurred. See [my other comment below](https://github.com/serokell/tzbot/pull/58#discussion_r1115529130).

test/Test/TzBot/RenderSpec.hs
dcastro, line 84: I also wanted to point out the inconsistent formats we're using here:

>           "02:00, Saturday, 11 March 2023 in America/Havana"
>          "_Warning! There is a time shift in America/Havana at 2023-03-12 05:00:00 UTC, so around this time conversion may depend on concrete date._"

In the first line, we're using `HH:MM, day of week, DD Month YY` and in the second we're using `YYYY-MM-DD HH:MM:SS`. Let's try to always use the first format for consistency.

YuriRomanowski: Great remark! I agree, let's elaborate to make more clear for user what we really want to say to him :)

YuriRomanowski: It looks like this:
изображение

YuriRomanowski: Now I use DD Month YY since I think time and day of week are not required.

test/Test/TzBot/RenderSpec.hs
```diff
@@ -119,6 +123,73 @@ test_renderSpec = TestGroup "Render"
...
       ]
     ]
+
+  , TestGroup "TimeshiftWarning"
+    [ testCase "Implicit date" $
+      mkChatCase nearTimeshift "10am" userMoscow userHavana
+      [ translWithCommonNote
+          "\"10am\", 11 March 2023 in Europe/Moscow"
+          "02:00, Saturday, 11 March 2023 in America/Havana"
+          "_Warning: We inferred that \"10am\" refers to 11 March 2023 in Europe/Moscow and converted it to America/Havana, but there is a time change near this date_:\n  \8226 _At 01:00, 12 March 2023 in America/Havana, the clocks are turned forward 1 hour(s)_.\n_Beware that if this inference is not correct and the sender meant a different date, the conversion may not be accurate._"
```
dcastro, line 133: Please use an interpolator and replace `\8226` with `•` to make this easier to read (in this test and the other ones where we use more complex multi-line strings)

```hs
          [int|s|  
            _Warning: We inferred that "10am" refers to 11 March 2023 in Europe/Moscow and converted it to America/Havana, but there is a time change near this date_:
              • _At 01:00, 12 March 2023 in America/Havana, the clocks are turned forward 1 hour(s)_.
            _Beware that if this inference is not correct and the sender meant a different date, the conversion may not be accurate._
          |]
```

dcastro, line 133: > _At 01:00, 12 March 2023 in America/Havana, the clocks are turned forward 1 hour(s)_

This seems incorrect now :sweat_smile: 

The clocks are not turned forward at 01:00, they're turned forward at 00:00: https://www.timeanddate.com/time/change/cuba/havana?year=2023

test/Test/TzBot/RenderSpec.hs
dcastro: Sweet, great work! :clap:

test/Test/TzBot/RenderSpec.hs
YuriRomanowski: Please let's just use plain strings because changing loads of well-formatted test entries is too painful.

test/Test/TzBot/RenderSpec.hs
dcastro: Yes, exactly :D

test/Test/TzBot/RenderSpec.hs
YuriRomanowski: You are definitely right, but when I started doing it that way, strings were way not so long, and now it's already too late. I guess it would be better to fix in a separate PR (maybe along with sections that you wanted to delete, I still remember! :laughing:)

YuriRomanowski: Ouch

src/TzBot/TimeReference.hs
```diff
@@ -143,24 +136,46 @@ timeReferenceToUTC sendersTZLabel eventTimestamp timeRef@TimeReference {..} =
     TZT.TZGap invalidTime _ _ -> TRTUInvalid $
       TimeShiftErrorInfo tzLabel (localDay invalidTime)
 
+  fromTZTimeWithFix :: Bool -> TZI.TZInfo -> LocalTime -> TZT.TZTime
+  fromTZTimeWithFix fixForward tzInfo localTime = case TZT.fromLocalTimeStrict tzInfo localTime of
+    Right res -> res
+    Left _ -> do
+      let diffHours = if fixForward then 3 else (-3)
+      let fix' = addLocalTime (TZT.hours diffHours)
```
dcastro, line 144: Why are we adding/subtracting 3 hours?

dcastro: It seems like simply using `TZT.fromLocalTime` on the "interval boundaries" (aka, the inferred date +/- 3 days) would be a good enough solution.

`TZT.fromLocalTime` applies a fair "adjustment" algorithm if the local time is invalid.

YuriRomanowski: If we get `TZError`, then it means we got into a timeshift, and we want to collect it for sure to emit a warning, that's why boundaries are adjusted this way (lower becomes lower, higher becomes higher)

YuriRomanowski: Well, sometimes `fromLocalTime` can leave a timeshift outside of the interval, that's why I made my own version.

dcastro: > Well, maybe that's not so important for us

Yea, I think so.

I mean, we want to perform the check in a `[-3 days, +3 days]` interval (aka, a 6 days interval).

But this is a completely arbitrary interval, there's nothing special about the number 6. We could just as easily have picked an 8 day interval.

So, if the boundaries of that interval happen to land exactly on top of a time shift, and `fromLocalTime` moves them a bit such that we now have a `[-2 days and 23hours, +3 days]` interval (aka, a 5 days and 23 hours interval), or a `[-3 days, +3 days and 1 hour]`, it's not a big deal, the margin is still quite wide.

src/TzBot/TimeReference.hs
```diff
@@ -144,23 +130,37 @@ timeReferenceToUTC sendersTZLabel eventTimestamp timeRef@TimeReference {..} =
...
+            if thatTimeOfCurrentDay >= eventLocalTime
+            then eventLocalTime
+            else TZT.addCalendarClip (TZT.calendarDays 1) eventLocalTime
+          checkDiff = 3
+          daysToCheckTimeshifts =
+            ( TZT.addCalendarClip (TZT.calendarDays (negate checkDiff)) res
+            , TZT.addCalendarClip (TZT.calendarDays checkDiff) res
+            )
+
+      (res, Just daysToCheckTimeshifts)
```
dcastro, line 148: This is getting really hard to read and review :/

Some examples of specific pain points are:

* `dayTransition` now mixes 2 different purposes: to calculate the date the user is referring to + to calculate the +/-3 days interval around that date. This should be split into two different functions, each with one well-defined purpose.
* Lack of documentation: for a reader who's new to the codebase, it's not easy to understand why we're calculating this interval in the first place.

I've gone ahead and wrote some docs for this feature [in this comment](https://github.com/serokell/tzbot/pull/63#discussion_r1116030513).

I also created a branch `diogo/warn-on-possible-timeshifts`, 041b86ba843c83368b0bf8a531b3a62556acbbf2, where I reverted `dayTransition` to its old definition, and extracted the "time shift" logic into a separate function `calculateTimeshiftCheckInterval`. Please have a look; if you like it, please copy those changes into this PR.

dcastro: Actually, I went one step further. I noticed the logic of this feature is split in 3 places:
* In `TimeReference.timeReferenceToUTC`, we decide which interval we're going to check for offset changes
* In `TZ.checkForTimeshifts`, we process this interval and actually check for offset changes
* In `Render.renderOnSuccess`, we call `TZ.checkForTimeshifts` and then render the result as a warning.

To avoid having this big split, I thought it would be better to unify the "decide which interval" and "process the interval" steps.

In other words, now `TZ.checkForTimeshifts` creates the interval and checks it for offset changes immediately.
`TimeReference.timeReferenceToUTC` does not care about this at all, it only cares about converting the time reference to UTC, and that's it.

I pushed another commit to my branch, 60c8020e02bcb3fbcb55d964cdd42f6c2c2e5afc, please have a look.

YuriRomanowski: Thank you very much! Looks good! (and also the ideal to which I should aspire :smile:)

dcastro (approved):

PR #59, opened 2023-02-05
[#53] Skip sending for matching timezones

### LINKED ISSUE #53: Skip sending messages when timezone is matching

From #4:

We don't always want to send an ephemeral message to all users. If a user meets one of the following criteria, we should skip it:
* if the user is a bot
* If timeRefLocation is a timezone, and if the user is in that same timezone.
* ~If timeRefLocation is an offset, and if, at the time of timeRefUTC, the user's timezone will be observing that same offset.~
        ~Note: use [timeZoneForUTCTime](https://hackage.haskell.org/package/tz-0.1.3.6/docs/Data-Time-Zones.html#v:timeZoneForUTCTime) to find out which offset the user will be observing at that time in their timezone.~

Let's discuss and handle the concern under this ticket.

At the moment we implemented a simpler strategy: never skip sending notification, by rather say "timezone is the same" when appropriate.

# Acceptance criteria

* We discussed the suggested strategy, accounted for possible corner cases.
* The selected solution is implemented.

YuriRomanowski:
## Description
Problem: Sometimes, when timezones in message and a receiver's one are the same, it's not required to actually translate anything because the original time reference is already relevant for the receiver.

Solution: for other users than sender, say "timezone is the same", for sender himself translate only time references that are in other timezones and ignore ones with his timezone, but translate everything when modal is triggered.

## Related issue(s)

Fixed #53 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/TzBot/ProcessEvents/Message.hs
+++ b/src/TzBot/ProcessEvents/Message.hs
@@ -26,7 +26,7 @@ import TzBot.Slack.API
 import TzBot.Slack.Events
 import TzBot.Slack.Fixtures qualified as Fixtures
 import TzBot.TimeReference (TimeReference(..))
-import TzBot.Util (isDevEnvironment, whenT, withMaybe)
+import TzBot.Util (whenT, withMaybe)
 
 data MessageEventType = METMessage | METMessageEdited
   deriving stock (Eq)
@@ -108,8 +108,7 @@ processMessageEvent' evt mEventType sender timeRefs =
   mbThreadId = mThreadId msg
 
   notBot u = not (uIsBot u)
-  notSameTimeZone u = uTz u /= uTz sender
-  notBotAndSameTimeZone u = notBot u && notSameTimeZone u
+  isSender userId = uId sender == userId
 
   getWhetherToShowHelpCmd :: BotM Bool
   getWhetherToShowHelpCmd = do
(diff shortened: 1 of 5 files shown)


src/TzBot/ProcessEvents/Message.hs
```diff
@@ -179,22 +178,20 @@ processMessageEvent' evt mEventType sender timeRefs =
...
-        [int|n|
-          Found invalid time references,
-          sending an ephemeral with them to the message sender
-          |]
-      sendAction mbPermalink asForSenderS errorsMsg (uId sender)
+    let ephemeralTemplate = renderTemplate asForMessageM now sender neTimeRefs
 
     let sendActionLocal userInChannelId = do
           userInChannel <- getUserCached userInChannelId
-          whenT (isDevEnvironment || notBotAndSameTimeZone userInChannel) do
```
dcastro, line 193: The `isDevEnvironment` function doesn't seem to be used anywhere anymore, could it be deleted from the `Util` module?

src/TzBot/ProcessEvents/Message.hs
```diff
@@ -179,22 +178,20 @@ processMessageEvent' evt mEventType sender timeRefs =
...
-            sendAction mbPermalink asForOthersS ephemeralMessage (uId userInChannel)
-            pure True
+          whenT (notBot userInChannel) do
+            let mbEphemeralMessage =
+                  renderAllTP userInChannel ephemeralTemplate
+            let senderFlag =
+                  if isSender $ uId userInChannel
+                  then asForSenderS
+                  else asForOthersS
+            withMaybe mbEphemeralMessage (pure False) \eph -> do
```
dcastro, line 192: Somewhat unrelated note, but: I think we could improve the docs in some places. E.g. in `ephemeralsMailing`, we should document when/why the `sendAction` argument function should return true or false.

dcastro, line 192: What does `eph` stand for?

dcastro: Oh nevermind, it stands for `ephemeral`

src/TzBot/Render.hs
```diff
@@ -130,18 +126,44 @@ renderSlackBlocks forSender =
...
+        Right _ -> Just renderedUserTime
+        Left refTzLabel ->
+          if refTzLabel /= userTzLabel
+          then Just renderedUserTime
+          else do
+            let isNotSender = ((/=) `on` uId) sender user
+                shouldShowThisTranslation = isNotSender || forModal
+            guard shouldShowThisTranslation
+            Just "your timezone is the same"
+  (<$> mbRenderedUserTime) $ \rt -> TranslationPair
```
dcastro, line 164:
```suggestion
  mbRenderedUserTime <&> \rt -> TranslationPair
```

src/TzBot/Render.hs
```diff
@@ -130,18 +126,44 @@ renderSlackBlocks forSender =
...
+      mbRenderedUserTime = case trsEithTzOffset of
+        Right _ -> Just renderedUserTime
+        Left refTzLabel ->
+          if refTzLabel /= userTzLabel
+          then Just renderedUserTime
+          else do
+            let isNotSender = ((/=) `on` uId) sender user
+                shouldShowThisTranslation = isNotSender || forModal
+            guard shouldShowThisTranslation
+            Just "your timezone is the same"
```
dcastro, line 163: I think a better way to phrase it might be "You're in this timezone"

PR #60, opened 2023-02-06
[Chore] Tzbot NixOS module

rvem:
## Description
1) Update nix infra to use flakes instead of `niv`
2) Provide NixOS module with tzbot systemd service

## Related issue(s)
https://issues.serokell.io/issue/OPS-1409

Fixed #

- Tests
        silently reappearing again.

- Documentation
    - Haddock


PhilTaken (approved):
not sure if I personally would leave the whole module in the flake but LGTM

rvem:
> not sure if I personally would leave the whole module in the flake but LGTM

Probably worth extracting in a separate file. But in general, I think it's easier to keep modules in the same repository to ease the migration between servers

PR #61, opened 2023-02-06
Improve parser

YuriRomanowski:
## Description

Further parser improvements.
Changes:

1. Add parsing of year, allow arbitrary ordering of time reference parts;
2. Add more time formats (according to the [table](https://www.notion.so/serokell/Requirements-7dc47068a206498989c4b899a767d208) here;
3. Add date formats.

## Related issue(s)

Continuation of #1 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/TzBot/Parser.hs
+++ b/src/TzBot/Parser.hs
@@ -14,7 +14,7 @@ import Data.String.Conversions (cs)
 import Data.Text qualified as T
 import Data.Text.Metrics (damerauLevenshteinNorm)
 import Data.Time (DayOfWeek(..))
-import Data.Time.Calendar.Compat (DayOfMonth, MonthOfYear)
+import Data.Time.Calendar.Compat (DayOfMonth, MonthOfYear, Year)
 import Data.Time.LocalTime (TimeOfDay(..))
 import Data.Time.Zones.All (TZLabel, tzNameLabelMap)
 import Glider.NLP.Tokenizer (Token(..), tokenize)
@@ -22,6 +22,7 @@ import Text.Megaparsec hiding (Token)
 
 import TzBot.Instances ()
 import TzBot.TimeReference
+import TzBot.Util
 import TzBot.Util qualified as CI
 
 type TzParser = Parsec Void [Token]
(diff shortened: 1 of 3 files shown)


dcastro:
Great work! :partying_face:  This actually fixes 2 issues I had noticed, but hadn't created a github issue yet:

* `01:30am in America/Winnipeg on 6th March` is parsed correctly, but if you reverse the order of the components, i.e. `01:30am on 6th March in America/Winnipeg`, then only the time and date were parsed. The timezone was ignored.
* `1am on 6th November 2022` - the year was being ignored

This ended up also fixing the issue @Martoon-00 described at the [bottom of this page](https://www.notion.so/serokell/Testing-iteration-1-from-dbfd12f048824fd68057f664fead248f), his example seems to work correctly now:

image

src/TzBot/TimeReference.hs
```diff
@@ -49,7 +47,7 @@ data TimeReference = TimeReference
 data DateReference
   = DaysFromToday Int
   | DayOfWeekRef DayOfWeek
-  | DayOfMonthRef DayOfMonth (Maybe MonthOfYear)
+  | DayOfMonthRef DayOfMonth (Maybe MonthOfYear) (Maybe Year)
```
dcastro, line 50: You can't have a `Year` without also providing a `MonthOfYear` (e.g., "1am on the 5th 2022" is not an acceptable input), so maybe the definition should be more like this:

```suggestion
  | DayOfMonthRef DayOfMonth (Maybe (MonthOfYear (Maybe Year)))
```

YuriRomanowski: Yeah, I thought about it, your version is definitely more correct, but it was too ugly and I decided to leave current version :smile: But you are right, let's fix it eventually

src/TzBot/Parser.hs
```diff
@@ -158,13 +159,52 @@ type TzParser = Parsec Void [Token]
...
+>>> parseTimeRefs "7:30pm 2022/08/3"
+[TimeReference {trText = "7:30pm 2022/08/3", trTimeOfDay = 19:30:00, trDateRef = Just (DayOfMonthRef 3 (Just (8,Just 2022))), trLocationRef = Nothing}]
+
+>>> parseTimeRefs "2022.8.03 7:30 pm "
+[TimeReference {trText = "2022.8.03 7:30 pm", trTimeOfDay = 19:30:00, trDateRef = Just (DayOfMonthRef 3 (Just (8,Just 2022))), trLocationRef = Nothing}]
+
+>>> parseTimeRefs "7:30pm 2022.8.03 America/Havana"
+[TimeReference {trText = "7:30pm 2022.8.03 America/Havana", trTimeOfDay = 19:30:00, trDateRef = Just (DayOfMonthRef 3 (Just (8,Just 2022))), trLocationRef = Just (TimeZoneRef America__Havana)}]
+
 -}
```
dcastro, line 201: While we're here, what do you think of also adding support for spaces in between offsets, e.g. `UTC +3`? Seems to be something users might write, and `timely` seems to support it too.

YuriRomanowski: Sure, let's support this too!

dcastro (approved):

PR #69, opened 2023-02-14
[#68] Flush feedback to the file immediately

### LINKED ISSUE #68: Flush feedback to the file immediately

Right now, we're using [`withFile`](https://hackage.haskell.org/package/universum-1.8.1/docs/Universum-Lifted-File.html#v:withFile) to open a file handle to `feedback.log`. This seems to use `BlockBuffering` by default, which means feedback is not immediately written to the log file. This might take some time to be written to file or, worst case scenario, is only written when the application shuts down.

We should use line buffering instead, so that output is flushed to the file as soon as a newline character is emitted.

# Acceptance criteria

* Feedback is written to the file immediately.

dcastro:
## Description

Problem: Right now, we're using `withFile` to open a file handle to `feedback.log`. This seems to use `BlockBuffering` by default, which means feedback is not immediately written to the log file. This might take some time to be written to file or, worst case scenario, is only written when the application shuts down.

Solution: Use line buffering instead, so that output is flushed to the file as soon as a newline character is emitted.

Also took the opportunity to do a couple of small fixes: replace a usage of `withObject ""` with `withObject "MessageEvent"`, and pull dependencies from hackage instead of github repos.

## Related issue(s)

Fixed #68

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/TzBot/BotMain.hs
+++ b/src/TzBot/BotMain.hs
@@ -15,6 +15,7 @@ import Slacker
   (defaultSlackConfig, handleThreadExceptionSensibly, runSocketMode, setApiToken, setAppToken,
   setGracefulShutdownHandler, setOnException)
 import System.Directory (doesFileExist)
+import System.IO (BufferMode(..), hSetBuffering)
 import Text.Interpolation.Nyan (int, rmode')
 import Time (hour)
 
@@ -104,4 +105,7 @@ withFeedbackConfig Config {..} action = do
     withFeedbackFile :: Maybe FilePath -> (Maybe Handle -> IO a) -> IO a
     withFeedbackFile mbPath action =
       withMaybe mbPath (action Nothing) $ \path ->
-        withFile path AppendMode (action . Just)
+        withFile path AppendMode \handle -> do
+          -- Use `LineBuffering` so that feedback is flushed to the file immediately.
+          hSetBuffering handle LineBuffering
+          action $ Just handle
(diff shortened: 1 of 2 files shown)


YuriRomanowski:
> pull dependencies from hackage instead of github repos.
There were some troubles with `tztime` and `doctest-parallel`. The former is not available inside CI for some reason (as your CI build says), the latter has an old version in our stack resolver, it doesn't work well with CI, while new versions play well with CI. Maybe I should have left some comments about why I used direct github links, but I thought it's not so critical.

dcastro:
> The former is not available inside CI for some reason (as your CI build says)

Oh I see. I updated some nix dependencies (`niv update hackage.nix && niv update haskell.nix && niv update nixpkgs`).

The more recent nixpkgs seems to be using a newer version of stack2cabal/hpaxk, so I had to update that too.

> the latter has an old version in our stack resolver, it doesn't work well with CI, while new versions play well with CI because Nix support was added; also the older version even has no logging so it's nearly impossible to debug when something is going wrong.

Right, I pinned the version to 0.3.0 in stack.yaml then.

The first build after this upgrade is taking a looong time to complete, but I'm fairly confident it'll work. Let's wait and see :)

dcastro:
Argh, I managed to get the "build" and "validate cabal files" steps [working](https://buildkite.com/serokell/tzbot/builds/339#0186507a-afb1-43d2-b582-2d3022973e4d), but hit a snag building some test suite and weeder. 

I'll just drop these commits from this PR and take care of that later.

YuriRomanowski (approved):

PR #75, opened 2023-02-16
[#70] Warn users on unrecognized timezone abbreviation

### LINKED ISSUE #70: Warn the sender when an unrecognized timezone abbreviation is used

The bot supports time references with time zone abbreviations, e.g. `5am in CST`.

However, as noted in [`timezone_abbreviations.md`](https://github.com/serokell/tzbot/blob/635077e77d6c85ab9c1711076aa1d5970444e7aa/docs/timezone_abbreviations.md), time zone abbreviations (1) are not standardized and (2) can be ambiguous.

> For example, there are five different interpretations of CST:
>
>   * Central Standard Time (North America)
>   * Central Standard Time (Australia)
>   * Central Summer Time (Australia)
>   * China Standard Time
>   * Cuba Standard Time

Nevertheless, we decided to use a subset of commonly used abbreviations.

However, we should bear in mind that the user might attempt to use an abbreviation we do not support.
For this reason, the document above also specifies how the bot is meant to behave when an unrecognized abbreviation is used:

> If a user sends a message with a reference to a timezone abbreviation that we don't support / doesn't exist, the bot notifies them of this.

And indeed, issue #1 where the parser was implemented also states:

> Timezone abbreviation. For simplicity, we should consider any all-uppercase 2-5 letter word a timezone abbreviation (except AM and PM, of course). Validation of the timezone abbreviation will happen in a separate issue.

However, it looks like the parser was in fact implemented by silently ignoring unrecognized timezone abbreviations.
Let's fix that here.

# Acceptance criteria

* The sender is warned when they use an unrecognized time zone abbreviation.

YuriRomanowski:
## Description
Problem: Currently, unrecognized timezone abbreviation is simply ignored as it would be an unrelated word.

Solution: Consider all-capitalized words with 2-5 characters length as timezone abbreviations and warn users when they are not known by us; add hints when encountered abbreviation is similar to some of known ones.

## Related issue(s)

Fixed #70 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/TzBot/Parser.hs
+++ b/src/TzBot/Parser.hs
@@ -8,9 +8,11 @@ module TzBot.Parser
 
 import Universum hiding (many, toList, try)
 
+import Data.Char (isUpper)
 import Data.Map qualified as M
 import Data.String.Conversions (cs)
 import Data.Text qualified as T
+import Data.Text.Metrics (damerauLevenshteinNorm)
 import Data.Time (DayOfWeek(..))
 import Data.Time.Calendar.Compat (DayOfMonth, MonthOfYear)
 import Data.Time.LocalTime (TimeOfDay(..))
@@ -30,7 +32,7 @@ type TzParser = Parsec Void [Token]
 [TimeReference {trText = "tuesday at 10am", trTimeOfDay = 10:00:00, trDateRef = Just (DayOfWeekRef Tuesday), trLocationRef = Nothing}]
 
 >>> parseTimeRefs "i can do it at 3pm MDT"
-[TimeReference {trText = "at 3pm MDT", trTimeOfDay = 15:00:00, trDateRef = Nothing, trLocationRef = Just (TimeZoneAbbreviationRef "MDT")}]
+[TimeReference {trText = "at 3pm MDT", trTimeOfDay = 15:00:00, trDateRef = Nothing, trLocationRef = Just (TimeZoneAbbreviationRef (TimeZoneAbbreviationInfo {tzaiAbbreviation = "MDT", tzaiOffsetMinutes = -360, tzaiFullName = "Mountain Daylight Time (North America)"}))}]
 
 >>> parseTimeRefs "how about between 2pm and 3pm?"
 [TimeReference {trText = "2pm", trTimeOfDay = 14:00:00, trDateRef = Nothing, trLocationRef = Nothing},TimeReference {trText = "3pm", trTimeOfDay = 15:00:00, trDateRef = Nothing, trLocationRef = Nothing}]
@@ -75,7 +77,7 @@ type TzParser = Parsec Void [Token]
 [TimeReference {trText = "9am in europe/london", trTimeOfDay = 09:00:00, trDateRef = Nothing, trLocationRef = Just (TimeZoneRef Europe__London)}]
 
 >>> parseTimeRefs "2pm CST"
-[TimeReference {trText = "2pm CST", trTimeOfDay = 14:00:00, trDateRef = Nothing, trLocationRef = Just (TimeZoneAbbreviationRef "CST")}]
+[TimeReference {trText = "2pm CST", trTimeOfDay = 14:00:00, trDateRef = Nothing, trLocationRef = Just (TimeZoneAbbreviationRef (TimeZoneAbbreviationInfo {tzaiAbbreviation = "CST", tzaiOffsetMinutes = -360, tzaiFullName = "Central Standard Time (North America)"}))}]
 
 >>> parseTimeRefs "10am UTC+03:00"
 [TimeReference {trText = "10am UTC+03:00", trTimeOfDay = 10:00:00, trDateRef = Nothing, trLocationRef = Just (OffsetRef 180)}]
@@ -146,9 +148,6 @@ type TzParser = Parsec Void [Token]
 >>> parseTimeRefs "today ,10am"
 [TimeReference {trText = "today ,10am", trTimeOfDay = 10:00:00, trDateRef = Just (DaysFromToday 0), trLocationRef = Nothing}]
 
->>> parseTimeRefs "10am MSKC"
-[TimeReference {trText = "10am", trTimeOfDay = 10:00:00, trDateRef = Nothing, trLocationRef = Nothing}]
-
 >>> parseTimeRefs "10am aMeRiCa/Argentina/Buenos_Aires"
 [TimeReference {trText = "10am aMeRiCa/Argentina/Buenos_Aires", trTimeOfDay = 10:00:00, trDateRef = Nothing, trLocationRef = Just (TimeZoneRef America__Argentina__Buenos_Aires)}]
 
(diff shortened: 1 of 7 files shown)


src/TzBot/Parser.hs
```diff
@@ -325,16 +339,37 @@ offsetRefParser = do
...
+              norm > similarityThreshold
+        let candidates =
+              filter predicate . map tzaiAbbreviation $ knownTimeZoneAbbreviations'
+        pure $ UnknownTimeZoneAbbrev (TimeZoneAbbreviation abbr) candidates
+    ]
+  -- That's added to avoid interference with UTC+... and GMT+... syntax.
   notFollowedBy $ choice' $ map punct ['+', '-']
   pure res
 
+isPossibleTimezoneAbbrev :: Text -> Bool
```
dcastro, line 365:
```suggestion
-- | We consider any string with 2-5 upper case letters (other than @AM@ or @PM@) a time zone abbreviation.
isPossibleTimezoneAbbrev :: Text -> Bool
```

src/TzBot/Parser.hs
```diff
@@ -325,16 +339,37 @@ offsetRefParser = do
...
-  pure . OffsetRef $ Offset $ sign * minutesTotal
+  pure $ OffsetRef (Offset $ sign * minutesTotal) Nothing
 
 tzAbbrRefParser :: TzParser LocationReference
 tzAbbrRefParser = do
   abbr <- anyWord
-  res <- TimeZoneAbbreviationRef . tzaiAbbreviation <$>
-    withStorage knownTimeZoneAbbreviations abbr
+  let similarityThreshold = 0.65 :: Ratio Int
+  res <- choice'
```
dcastro, line 348:
```suggestion
  -- Check if this word is a recognized timezone abbreviation.
  --
  -- If it isn't, check whether the user likely meant to refer to some
  -- timezone abbreviation, but possibly made a typo or used an abbreviation we don't support.
  -- If so, and assuming it was a typo, suggest abbreviations similar enough to what the user entered.
  -- E.g., if they entered "WETS", they likely meant either "WET" or "WEST".
  res <- choice'
```

src/TzBot/Render.hs
```diff
@@ -219,12 +220,17 @@ renderEphemeralMessageTranslationPair modalFlag sender (timeRef, result) = case
...
-  TRTUInvalidTimeZoneAbbrev abbrev ->
+  TRTUInvalidTimeZoneAbbrev UnknownTimeZoneAbbrev {..} -> do
+    let mbNeCandidates = nonEmpty utzaCandidates
+        mbCandidatesNoteForSender = flip fmap mbNeCandidates \neCandidates -> do
+          let renderedCandidates =
+                T.intercalate ", " $
+                  map unTimeZoneAbbreviation $ NE.toList neCandidates
+          [int||_Maybe you meant: #{renderedCandidates}_|]
     Left $ TranslationPair
       (trText timeRef)
```
dcastro, line 231: Maybe we should also wrap the original text in quotes `""` like we do in the successful case? For coherence:

image

In the first block, `4am WETS` is not in quotes and in the second block `"5am WEST"` is.

dcastro:
Great idea to suggest similar known timezone abbreviations, well done! :smile:  :+1:

src/TzBot/TimeReference.hs
```diff
@@ -55,9 +53,9 @@ data DateReference
 data LocationReference
   = TimeZoneRef TZLabel
   -- ^ A timezone name, e.g. @Europe/London@.
-  | OffsetRef Offset
+  | OffsetRef Offset (Maybe TimeZoneAbbreviationInfo)
```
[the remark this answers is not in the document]
YuriRomanowski: Yes, you are definitely right that there is some redundancy, but from this `TimeZoneAbbreviationInfo` we want other fields like the description, not the offset. So, if it's `Nothing`, we know that explicit offset was used like `UTC+6`, and if it's `Just` we know that abbreviation was used, so we can distinguish these cases later.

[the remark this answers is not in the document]
YuriRomanowski: > I think it makes sense to keep the old | TimeZoneAbbreviationRef TimeZoneAbbreviation constructor

Currently, in the parser we firstly check for known abbreviations which may not satisfy our requirements that are stated in `isPossibleAbbrev` function, and in case of match we don't need the `TimeZoneAbbreviation` anymore, we can just use `TimeZoneAbbreviationInfo`. And if something looks like an abbreviation (2-5 length and all uppercase), but it's not known, we are going to report an error. These cases, even although they look similar, are in fact essentially different. So I merged the known-abbreviation case into `OffsetRef` case, and made a new constructor for the case unknown abbreviations. Does it make sense?

src/TzBot/TimeReference.hs
```diff
@@ -161,15 +164,13 @@ timeReferenceToUTC sendersTZLabel eventTimestamp TimeReference {..} =
   -- The outer `Either` acts like an error carrier, so for it we use `pure` and `<$>`,
   -- and the inner `Either` carries one of the possible equitable results, so
   -- for it we use `Right` or `Left`.
-  mbEitherTzOrOffset :: Either TimeZoneAbbreviation (Either (TZLabel, Bool) Offset)
+  mbEitherTzOrOffset
+    :: Either UnknownTimeZoneAbbrev (Either (TZLabel, Bool) (Offset, Maybe TimeZoneAbbreviationInfo))
```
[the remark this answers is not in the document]
YuriRomanowski: This will be used [here](https://github.com/serokell/tzbot/pull/76), so let's keep it as it is here please :)

src/TzBot/Parser.hs
YuriRomanowski: Isn't it obvious from the code? Seems like this haddock will double the code contents, in my mind docs should rather say "why", not "what".

YuriRomanowski: Also this function is not exported, so there will be no need to use this function outside of this module and hence no need to open extra source file to get an idea of what's this function is doing because of absent docs.

src/TzBot/Render.hs
YuriRomanowski: Good catch, thanks :+1:

src/TzBot/TimeReference.hs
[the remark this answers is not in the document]
dcastro: Gotcha! ;)

src/TzBot/Parser.hs
dcastro: My initial motivation was to save other developers from having to carefully inspect the function's definition to learn what it does - but you're right, this particular function is so small and easy to understand, this is not needed ^^

src/TzBot/TimeReference.hs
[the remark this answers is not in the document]
dcastro: > and in case of match we don't need the `TimeZoneAbbreviation` anymore, we can just use `TimeZoneAbbreviationInfo`. 

Oh sorry, I should have said `TimeZoneAbbreviationInfo` instead of `TimeZoneAbbreviation`.

But my main point was that I think it makes sense to keep a separate constructor for time zone abbreviations and another one for offsets.

So, instead of:
```hs
| OffsetRef Offset (Maybe TimeZoneAbbreviationInfo)
```

We'd have:

```hs
| OffsetRef Offset 
| TimeZoneAbbreviationRef TimeZoneAbbreviationInfo
```

And `tzAbbrRefParser` would construct a `TimeZoneAbbreviationRef/UnknownTimeZoneAbbreviationRef`, and `offsetRefParser` would construct a `OffsetRef`.

Or is this not possible for some reason I'm not seeing? :sweat_smile: 




And if something looks like an abbreviation (2-5 length and all uppercase), but it's not known, we are going to report an error. These cases, even although they look similar, are in fact essentially different. So I merged the known-abbreviation case into the `OffsetRef` constructor, and made a new constructor for the case unknown abbreviations. Does it make sense?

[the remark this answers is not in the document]
YuriRomanowski: Well, I think you are seeing right :+1:

dcastro (approved):

PR #77, opened 2023-02-16
[#66] Improve warnings about timeshifts

### LINKED ISSUE #66: Warnings about time shifts

When a user writes a message with an invalid/ambiguous time reference, we emit the warnings below. However, these messages could be a bit more helpful, and are sometimes even misleading.

----

```
"01:30am in America/Winnipeg on 5th november"
ambiguous because of the time shift

There is a timeshift in America/Winnipeg around the specified time and this particular timestamp can be possible with different offsets, please define the offset explicitly.
```

* "please define the offset explicitly"
  We know exactly which offsets are possible, the [`TZError.TZOVerlap`](https://hackage.haskell.org/package/tztime-0.1.0.0/docs/Data-Time-TZTime.html#t:TZError) constructor contains this information.
  We should tell the user something more specific so that they'll know _how_ to fix their message.

Suggestion:
```
"01:30am in America/Winnipeg on 5th november"
Ambiguous time

The clocks are turned backward <x> hour(s) (and <y> minute(s)) in America/Winnipeg around the specified time and this particular time occurs twice.
Please edit your message or write a new one and specify one of the following offsets: UTC-05:00 or UTC-06:00.
```

-------

```
"02:30am in America/Winnipeg on 12th march"
invalid because of the time shift

There is a timeshift in America/Winnipeg around the specified time and this particular timestamp does not exist, please define the offset explicitly.
```

* "timeshift" vs "time shift": we should use consistent spelling.

* "please define the offset explicitly"
  Defining the offset explicitly would not help. The time simply does not occur.

* We could use the times given by the `TZError.TZGap` constructor as suggestion for how to fix the problem. In this example, the constructor will return the given time (02:30) shifted backward and forward by the duration (1 hour) of the gap, i.e. 01:30 and 03:30.

Suggestion:

```
"02:30am in America/Winnipeg on 12th march"
Invalid time

The clocks are turned forward <x> hour(s) (and <y> minute(s)) in America/Winnipeg around the specified time and this particular time does not occur.
Please edit your message or write a new one and amend the time. Did you mean 01:30 or 03:30 instead?
```

# Acceptance criteria

* The "ambiguous" warning tells users which offsets are possible
* The "invalid" warning does not tell users to specify an offset, as that wouldn't help
* The "invalid" warning suggests some possible valid times to the user

YuriRomanowski:
## Description

Problem: Current messages about timeshifts can be more informative.

Solution: Add information about offsets and offset change.

## Related issue(s)

Fixed #66 

- Tests
        silently reappearing again.

- Documentation
    - Haddock


src/TzBot/TimeReference.hs
```diff
@@ -244,13 +264,26 @@ data TimeShiftErrorInfo = TimeShiftErrorInfo
...
 
+data OverlapInfo = OverlapInfo
+  { oiFirstOccurrence :: TZT.TZTime
+  , oiSecondOccurrence :: TZT.TZTime
+  , oiErrorInfo :: TimeShiftErrorInfo
+  } deriving stock (Eq, Show)
+
+data GapInfo = GapInfo
+  { giPreviousTime :: TZT.TZTime
+  , giNextTime :: TZT.TZTime
```
dcastro, line 275: Let's add some docs here because the contents of these 2 fields may not be immediately obvious to some readers.

```suggestion
    { giPreviousTime :: TZT.TZTime
    -- ^ The local time specified by the user shifted backward by the length of the gap.
    -- E.g., if they said "2:30am" but the clocks were turned forward 1 hour at 2am (such that 2:30am did not occur),
    -- then this field would be 2:30am - 1 hour = 1:30am.
    , giNextTime :: TZT.TZTime
    -- ^ The local time specified by the user shifted forward by the length of the gap.
    -- E.g., if they said "2:30am" but the clocks were turned forward 1 hour at 2am (such that 2:30am did not occur),
    -- then this field would be 2:30am + 1 hour = 3:30am.
```

src/TzBot/Render.hs
```diff
@@ -167,18 +180,72 @@ renderOnSuccess (ModalFlag forModal) sender timeRef TimeRefSuccess {..} user = d
...
+        The clocks are turned backward #{hoursDiff} hour(s)#{shownMinutesDiff}
+        in #{shownTZ forSender} around the specified time and this
+        particular time occurs twice with offsets #{firstOffset} and #{secondOffset}.
+        |] :: Builder
+      additionForSender = [int|n|
+        Please edit your message or write a new one and specify an offset explicitly.
+        |] :: Builder
+  TranslationPair
+    { tuTimeRef = getOriginalTimeRef sender timeRef tseiOriginalDate
+    , tuTranslation = "Ambiguous because of the time shift"
```
dcastro, line 204: I think we should use the article "a" instead of "the" here.

But given that we already have such a detailed explanation below, perhaps simply saying "Ambiguous time" (or "Invalid time" in the case of a gap) would be enough.

test/Test/TzBot/RenderSpec.hs
```diff
@@ -90,6 +90,46 @@ test_renderSpec = TestGroup "Render"
...
+          }
+        ]
+      ]
+    , TestGroup "Gap"
+      [ testCase "Explicit timezone" $
+        mkChatCase arbitraryTime1 "0:30 in america/havana on the 12th march" userHavana userMoscow
+        [ TranslationPair
+          { tuTimeRef = "\"0:30 in america/havana on the 12th march\""
+          , tuTranslation = "Invalid because of the time shift"
+          , tuNoteForSender = Just "_The clocks are turned forward 1 hour(s) from UTC-05:00 to UTC-04:00 in America/Havana around the specified time and this particular time does not occur. Please edit your message or write a new one and specify an offset explicitly._"
```
dcastro, line 119: I thought we weren't going to show the offsets in the case of a gap; we were going to show the boundaries of the gap and offer `giPreviousTime` and `giNextTime` as possible suggestions, like in [this example](https://github.com/serokell/tzbot/issues/66#issuecomment-1434284763) here, right?

```
"02:30am in America/Winnipeg on 12th march"
Invalid time

At 02:00, the clocks are turned forward <x> hour(s) (and <y> minute(s)) to 03:00 and this particular time does not occur in America/Winnipeg.
Please edit your message or write a new one and amend the time. Did you mean 01:30 or 03:30 instead?
```

src/TzBot/Render.hs
YuriRomanowski: Oh, articles, my infinite pain :disappointed:

test/Test/TzBot/RenderSpec.hs
YuriRomanowski: Yep :smile:

dcastro: Great work :D

test/Test/TzBot/RenderSpec.hs
```diff
@@ -114,6 +114,46 @@ test_renderSpec = TestGroup "Render"
...
+          , tuNoteForOthers = Just "_The clocks are turned backward 1 hour(s) in America/Havana around the specified time and this particular time occurs twice with offsets UTC-04:00 and UTC-05:00._"
+          }
+        ]
+      , testCase "Implicit timezone" $
+        mkChatCase arbitraryTime1 "0:30 on the 6th november" userHavana userMoscow
+        [ TranslationPair
+          { tuTimeRef = "\"0:30 on the 6th november\" in America/Havana"
+          , tuTranslation = "Ambiguous time"
+          , tuNoteForSender = Just "_The clocks are turned backward 1 hour(s) in your timezone (America/Havana) around the specified time and this particular time occurs twice with offsets UTC-04:00 and UTC-05:00. Please edit your message or write a new one and specify an offset explicitly._"
+          , tuNoteForOthers = Just "_The clocks are turned backward 1 hour(s) in the sender's timezone (America/Havana) around the specified time and this particular time occurs twice with offsets UTC-04:00 and UTC-05:00._"
```
dcastro, line 133: Great work with diving into `tz`'s internals and finding out the price time at which the offset changed!

Since we used that information to improve the "invalid time" message, I think we should also use it to improve the "ambiguous time" message.

Instead of "The clocks are turned ... _around the specific time_ ...", I think we should use the specific time and say something like:

> At hh:mm, the clocks are turned backward 1 hour(s) to hh:mm and this particular time occurs twice in America/Winnipeg, first with the offset UTC-05:00 and then with UTC-06:00.
Please edit your message or write a new one and specify an offset explicitly.

src/TzBot/Render.hs
```diff
@@ -244,6 +245,91 @@ getOriginalTimeRef sender timeRef originalDay = do
...
+  let TimeShiftErrorInfo {..} = giErrorInfo
+      prevOffset = tztimeOffset giPreviousTime
+      nextOffset = tztimeOffset giNextTime
+      (hoursDiff, mbMinutesDiff) = getAbsOffsetDiff prevOffset nextOffset
+
+      shownMinutesDiff = mbMinutesDiff <&> \d ->
+        [int|| (and #{d} minutes)|] :: Builder
+
+      isImplicitSenderTimezone = isNothing $ trLocationRef timeRef
+      timeshift = findLastTimeshift tseiRefTimeZone $ TZT.toUTC giNextTime
```
dcastro, line 303: > findLastTimeshift tseiRefTimeZone $ TZT.toUTC giNextTime

* We have two important variables in scope here, `tseiRefTimeZone :: TZLabel` and `giNextTime :: TZTime`
* We're passing `tseiRefTimeZone :: TZLabel` into `findLastTimeshift`, which then immediately converts it to a `TZInfo`
* We're converting `giNextTime :: TZTime` to `UTCTime` and passing it to `findLastTimeshift`

I noticed we have a `TZLabel` in scope, but what we really need is a `TZInfo`; and that `TZInfo` is already stored inside `giNextTime`.

So my conclusion is that:
* `giNextTime` already contains all the information we need, we can just pass that straight into `findLastTimeshift`
* the `TimeShiftErrorInfo.tseiRefTimeZone` field is redundant and we can just delete it.

I think we can change the definition of `findLastTimeshift` from this:

```hs
findLastTimeshift :: TZLabel -> UTCTime -> TimeShift
findLastTimeshift tsTzInfo (utcTimeToInt64 -> after) = do
  ...
```

to this:

```hs
findLastTimeshift :: TZTime -> TimeShift
findLastTimeshift tzt = do
  let tsTzInfo = TZT.tzTimeTZInfo tzt
      after = utcTimeToInt64 $ TZT.toUTC tzt
  ...
```

stack.yaml
```diff
@@ -7,6 +7,8 @@ resolver: lts-19.23
 packages:
 - .
 
+system-ghc: true
```
dcastro, line 10: Did you commit this by mistake?

YuriRomanowski: Yep :(

src/TzBot/Render.hs
dcastro: Cool, looks like this is now done, I think we just need to delete the `TimeShiftErrorInfo.tseiRefTimeZone` field now (unless there's another reason why it's needed)

stack-hie.yaml.lock
```diff
@@ -0,0 +1,99 @@
+# This file was autogenerated by Stack.
```
dcastro, line 1: Think you committed this file by accident as well :sweat_smile:

dcastro (approved):
Great work! :clap:

src/TzBot/Render.hs
YuriRomanowski: It seems that the best way is just pass full `TZError` to `findLastTimeshift` :smile: And rename it to `getTimeshift`

PR #78, opened 2023-02-17
[#67] Refactor timeReferenceToUTC

### LINKED ISSUE #67: Refactor `timeReferenceToUTC`
When a time reference contains a static offset (e.g. UTC+05:00), `timeReferenceToUTC` uses this auxiliary `convertOffsetTimeToUtc` function to convert the event timestamp in UTC to the given offset.

https://github.com/serokell/tzbot/blob/635077e77d6c85ab9c1711076aa1d5970444e7aa/src/TzBot/TimeReference.hs#L87-L89

And then later we use `convertOffsetTimeToUtc` to convert back.

However, `LocalTime` is not the best type for this situation. `LocalTime` is used to define a time in some _unknown_ offset. Here, we do know the offset, it's UTC+05:00, so we should use `ZonedTime` instead (which is basically a `LocalTime` + an offset).

-----

Additional improvement:

Better yet, we could return a `TZTime`, and construct it using a `TZInfo` composed of a single static offset.

For example, to construct a `TZInfo` that reprents the static offset UTC+05:00, we could do:

```hs
offsetInfo = TZInfo "+05" (TZ (singleton minBound) (singleton (5 * 60))) (singleton (False, "+05"))
```

And then convert the timestamp from UTC to this offset with:

```hs
TZT.fromUTC offsetInfo eventTimestamp
```

By making this change, we could then unify the [`Right offset` branch](https://github.com/serokell/tzbot/blob/635077e77d6c85ab9c1711076aa1d5970444e7aa/src/TzBot/TimeReference.hs#L112) and then [`Right (Left (tzLabel, implicitSenderTimezone))` branch](https://github.com/serokell/tzbot/blob/635077e77d6c85ab9c1711076aa1d5970444e7aa/src/TzBot/TimeReference.hs#L123).

They use pretty much the same conversion logic, exception one uses a `TZInfo` from a static offset and the other a `TZInfo` for an actual time zone.

# Acceptance criteria

* The conversion logic in the two branches of `timeReferenceToUTC` is unified.

YuriRomanowski:
## Description

Problem: We use two different code branches for cases of static offset and timezone.

Solution: Unify two branches in timeReferenceToUTC.

## Related issue(s)

Fixed #67 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/TzBot/Render.hs
+++ b/src/TzBot/Render.hs
@@ -152,9 +152,9 @@ renderOnSuccess (ModalFlag forModal) sender timeRef TimeRefSuccess {..} user = d
       renderedUserTime = do
         let q = renderUserTime userTzLabel trsUtcResult
         [int||#{q} in #{userTzLabel}|] :: Text
-      mbRenderedUserTime = case trsEithTzOffset of
-        Right _ -> Just renderedUserTime
-        Left refTzLabel ->
+      mbRenderedUserTime = case getTzLabelMaybe (uTz sender) timeRef of
+        Nothing -> Just renderedUserTime
+        Just refTzLabel ->
           if refTzLabel /= userTzLabel
           then Just renderedUserTime
           else do
@@ -196,7 +196,7 @@ renderEphemeralMessageTranslationPair modalFlag sender (timeRef, result) = case 
   TRTUSuccess timeRefSuc ->
     Right $ renderOnSuccess modalFlag sender timeRef timeRefSuc
   TRTUAmbiguous TimeShiftErrorInfo {..} -> do
-    let shownTZ = shownTimezone tseiIsImplicitSenderTimezone tseiRefTimeZone
+    let shownTZ = shownTimezone tseiRefTimeZone
     Left $ TranslationPair
       { tuTimeRef = getOriginalTimeRef sender timeRef tseiOriginalDate
       , tuTranslation = "Ambiguous because of the time shift"
@@ -210,7 +210,7 @@ renderEphemeralMessageTranslationPair modalFlag sender (timeRef, result) = case 
              different offsets._|]
       }
   TRTUInvalid TimeShiftErrorInfo {..} -> do
-    let shownTZ = shownTimezone tseiIsImplicitSenderTimezone tseiRefTimeZone
+    let shownTZ = shownTimezone tseiRefTimeZone
     Left $ TranslationPair
       { tuTimeRef = getOriginalTimeRef sender timeRef tseiOriginalDate
       , tuTranslation = "Invalid because of the time shift"
@@ -236,13 +236,15 @@ renderEphemeralMessageTranslationPair modalFlag sender (timeRef, result) = case 
       mbCandidatesNoteForSender
       Nothing
   where
-    shownTimezone :: Bool -> TZLabel -> Bool -> Builder
-    shownTimezone implicitSenderTimezone tzLabel forSender
+    shownTimezone :: TZLabel -> Bool -> Builder
+    shownTimezone tzLabel forSender
       | implicitSenderTimezone =
           if forSender
           then [int||your timezone (#{tzLabel})|]
           else [int||the sender's timezone (#{tzLabel})|]
       | otherwise = [int||#{tzLabel}|]
+      where
+        implicitSenderTimezone = isNothing $ trLocationRef timeRef
 
 renderUserTime :: TZLabel -> UTCTime -> String
 renderUserTime tzLabel refTime = do
--- a/src/TzBot/TimeReference.hs
+++ b/src/TzBot/TimeReference.hs
@@ -9,14 +9,18 @@ import Universum
 import Control.Arrow ((>>>))
 import Data.List.NonEmpty qualified as NE
 import Data.Time
-  (Day, DayOfWeek, LocalTime(localDay), NominalDiffTime, TimeOfDay, TimeZone, UTCTime, addLocalTime,
-  diffLocalTime, localTimeToUTC, nominalDay, toGregorian, utc, utcToLocalTime)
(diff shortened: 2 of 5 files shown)


src/TzBot/TimeReference.hs
```diff
@@ -107,40 +92,33 @@ timeReferenceToUTC
...
+      let eventLocalTime = TZT.fromUTC tzInfo eventTimestamp
       let eithRefTime = eventLocalTime & TZT.modifyLocalStrict (
             dayTransition >>> TZT.atTimeOfDay trTimeOfDay
             )
       case eithRefTime of
-        Left err -> tzErrorToResult implicitSenderTimezone tzLabel err
+        Left err -> do
+          let mbRefTzLabel = case trLocationRef of
+                Just (TimeZoneRef refTzLabel) -> Just refTzLabel
+                Nothing -> Just sendersTZLabel
```
dcastro, line 106: Looks like the logic of "if the user did not specify a time zone, default to using their current time zone" is now repeated in 3 places:

1. Here, line 106
2. In `eithTzInfo`, line 144
3. In `TzBot.Render.renderOnSuccess`, line 156

Maybe it would be useful to encapsulate this in one single function, e.g. `getTZLabelMaybe :: TimeReference -> Maybe TZLabel`?

src/TzBot/TimeReference.hs
```diff
@@ -107,40 +92,33 @@ timeReferenceToUTC
...
+                _ -> Nothing
+          case mbRefTzLabel of
+            Nothing -> error "impossible happened: got TZError for a static offset"
```
dcastro, line 109: We don't really need this `error` branch if we change `TimeShiftErrorInfo.tseiRefTimeZone` from a `TZLabel` to a `TZInfo`.

Maybe I'm not explaining myself correctly; I pushed a demo of what I mean to `diogo/#67-refactor-timeReferenceToUTC`, see commit `991a597c1d907e3ebbcd76147eeb1e33abd1fb5a`

YuriRomanowski: Well, I understood you correctly, but preferred to emit `error` explicitly rather than silently pass artificial `TZInfo` to the rendering stage, because in fact two behaviors are equivalent, but explicit `error` makes the code just more readable I believe.

src/TzBot/TimeReference.hs
```diff
@@ -161,16 +139,22 @@ timeReferenceToUTC sendersTZLabel eventTimestamp TimeReference {..} =
   -- The outer `Either` acts like an error carrier, so for it we use `pure` and `<$>`,
   -- and the inner `Either` carries one of the possible equitable results, so
   -- for it we use `Right` or `Left`.
```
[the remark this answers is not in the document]
YuriRomanowski: Or deleted :)

src/TzBot/TimeReference.hs
```diff
@@ -161,16 +139,22 @@ timeReferenceToUTC sendersTZLabel eventTimestamp TimeReference {..} =
...
+      tzInfoFromOffset . tzaiOffsetMinutes <$>
           maybe (Left abbrev) pure
             (CI.lookup (unTimeZoneAbbreviation abbrev) knownTimeZoneAbbreviations)
 
+  tzInfoFromOffset :: Offset -> TZI.TZInfo
+  tzInfoFromOffset (Offset offsetMinutes) =
+    TZI.TZInfo "" $ TZ
+        (VU.singleton minBound)
+        (VU.singleton $ secondsPerMinute * offsetMinutes)
+        (V.singleton (False, ""))
```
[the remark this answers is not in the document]
YuriRomanowski: I have picked ready `Buildable Offset` from another PR :)

src/TzBot/TimeReference.hs
YuriRomanowski: The second one you mentioned is slightly different from others, but `getTzLabelMaybe` replaced 1 and 3 quite well

src/TzBot/TimeReference.hs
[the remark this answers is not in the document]
dcastro: Well done ;)

src/TzBot/TimeReference.hs
```diff
@@ -128,41 +120,31 @@ timeReferenceToUTC
...
+    Right tzInfo -> do
+      let eventLocalTime = TZT.fromUTC tzInfo eventTimestamp
       let eithRefTime = eventLocalTime & TZT.modifyLocalStrict (
             dayTransition >>> TZT.atTimeOfDay trTimeOfDay
             )
       case eithRefTime of
-        Left err -> tzErrorToResult implicitSenderTimezone tzLabel err
+        Left err -> do
+          let mbRefTzLabel = getTzLabelMaybe sendersTZLabel timeRef
+          case mbRefTzLabel of
```
dcastro, line 134: Minor suggestion, to avoid having too many variable names floating around :sweat_smile: 

```suggestion
          case getTzLabelMaybe sendersTZLabel timeRef of
```

src/TzBot/Render.hs
```diff
@@ -152,9 +152,10 @@ renderOnSuccess (ModalFlag forModal) sender timeRef TimeRefSuccess {..} user = d
       renderedUserTime = do
         let q = renderUserTime userTzLabel trsUtcResult
         [int||#{q} in #{userTzLabel}|] :: Text
-      mbRenderedUserTime = case trsEithTzOffset of
-        Right _ -> Just renderedUserTime
-        Left refTzLabel ->
+      mbRefTzLabel = getTzLabelMaybe (uTz sender) timeRef
+      mbRenderedUserTime = case mbRefTzLabel of
```
dcastro, line 156:
```suggestion
      mbRenderedUserTime = case getTzLabelMaybe (uTz sender) timeRef of
```

dcastro (approved):

PR #80, opened 2023-02-17
[#79] Add support for hyphenated intervals

### LINKED ISSUE #16: Parsing multiple time mentions that share the same parts
~~Currently, we seem to handle only one mention of time per message~~

## Implicit time parts problem

There seems to be an important nuance in the case when the message contains multiple time entries, imagine this text:

```
Let's go on Wednesday at 10:00 or 11:00.
```

Here `Wednesday` applies both to `10:00` and `11:00`, and we should parse it accordingly.

---

Another example:

```
How about Wednesday at 10:00 / 11:00 OR 14:00 / 15:00 at Thursday
```

This should parse as

```
How about (Wednesday at 10:00 / 11:00) OR (14:00 / 15:00 at Thursday)
```

On the one hand, formalizing such rules sounds like a pretty tricky task.
On the other, sometimes such messages can be ambiguous for humans too, and ambiguity leads to miscommunication and wasted time, so maybe forcing some rules would not be that bad.

---

Btw, a common example of ambiguity: `11:00` used by people who live with 24h format would be ambigous for someone who got used to `am`/`pm` format.

I think, in such case we really should suggest the user to edit their message.
So having failing parsers is not a bad thing.

---

Finally, another bunch of examples: 

```
1. How about 10:00 or 11:00 am today?    // Does "am" apply to 10:00?
2. How about 13:00 or 11:00 am today?    // What here? 
3. How about 13:00 or 11:00 today?       // Can we infer that 11:00 is in 24h format?
```

# Acceptance criteria

* We have investigated whether supporting multiple time entries in a message makes sense.
* We wrote down a **real ton** of tests collected by at least 3 people.
* The Slack bot behaves accordingly to those tests.

### LINKED ISSUE #79: Add support for simple hyphenated intervals
There a lot of situations when time contexts are spread over different sentences and even messages, some of them we've listed in the issue report https://github.com/serokell/tzbot/issues/16. Let's add here at least for the very simple and widespread variation - hyphenated intervals.

# Acceptance criteria
* Simple hyphenated intervals are handled within common context

This issue is a part of https://github.com/serokell/tzbot/issues/16

YuriRomanowski:
## Description

Problem: If a user writes "I'll be available around 10am-11am UTC", 10am has no location reference.

Solution: Add support for simple hyphenated and slashed intervals sharing all other context, resulting two different time references that are processed separately further. Also added conjugations via "and" and "or".

## Related issue(s)

Fixed #79, partly fixed #16

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/TzBot/Parser.hs
+++ b/src/TzBot/Parser.hs
@@ -9,6 +9,7 @@ module TzBot.Parser
 import Universum hiding (many, toList, try)
 
 import Data.Char (isUpper)
+import Data.List qualified as L
 import Data.Map qualified as M
 import Data.String.Conversions (cs)
 import Data.Text qualified as T
@@ -18,6 +19,7 @@ import Data.Time.Calendar.Compat (DayOfMonth, MonthOfYear, Year)
 import Data.Time.LocalTime (TimeOfDay(..))
 import Data.Time.Zones.All (TZLabel, tzNameLabelMap)
 import Glider.NLP.Tokenizer (Token(..), tokenize)
+import Text.Interpolation.Nyan (int, rmode')
 import Text.Megaparsec hiding (Token)
 
 import TzBot.Instances ()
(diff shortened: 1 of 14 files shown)


YuriRomanowski:
When an interval is encountered, it's split into two separate time references, which are then processed independently, and as you can see on the screen, it can lead to different dates inferred for two ends of the interval. We could share their context even inside `timeReferenceToUTC` functions, but it seems that really a lot of refactoring is required, and results are still very poor because really a lot of examples are handled wrongly. Maybe we need to handle it later in some completely different way (along with #16)
изображение

src/TzBot/Parser.hs
```diff
@@ -271,6 +288,33 @@ builderParser allowSpace b = do
...
 
+data TimeOfDayWithText = TimeOfDayWithText
+  { todwtText :: [Token]
+  , todwtTime :: TimeOfDay
+  }
+
+data TimeEntry
+  = TESingle TimeOfDayWithText
+  | TEInterval TimeOfDayWithText TimeOfDayWithText
+  | TESlash TimeOfDayWithText TimeOfDayWithText
```
dcastro, line 299: Do we need both the `TEInterval` and `TESlash` constructors?

When we match on them (lines 259 and 260), we treat the same way, so looks like `TEInterval` would be enough.

src/TzBot/Parser.hs
```diff
@@ -216,31 +225,39 @@ timeRefsParser = choice'
...
   ]
 
 -- | Parses a single 'TimeReference', consuming all input.
-timeRefParser :: TzParser TimeReference
+timeRefParser :: TzParser [TimeReference]
 timeRefParser = do
   _ <- space
-  (newTrText, timeReference) <- match timeRefParser'
-  return timeReference { trText = concatTokens newTrText }
+  timeRefParser'
```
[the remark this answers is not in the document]
dcastro: BTW, why is this space mandatory and not optional? I see this line was introduced in 834067c2e5bc237774eb37e5cdecae43de691d91, but it's not clear why the space is not optional. 

Can you please leave a comment here documenting the reason?

src/TzBot/Parser.hs
dcastro: If we do this, then the `delimitedPair` local function won't need a `bldr` parameter.

src/TzBot/Parser.hs
```diff
@@ -271,6 +288,33 @@ builderParser allowSpace b = do
...
     Nothing -> pure b
 
+data TimeOfDayWithText = TimeOfDayWithText
+  { todwtText :: [Token]
+  , todwtTime :: TimeOfDay
+  }
+
+data TimeEntry
+  = TESingle TimeOfDayWithText
+  | TEInterval TimeOfDayWithText TimeOfDayWithText
```
dcastro, line 298: Let's also add some haddocks to illustrate the point of this data type:

```suggestion
  = TESingle TimeOfDayWithText
  -- ^ E.g. @10am@
  | TEInterval TimeOfDayWithText TimeOfDayWithText
  -- ^ E.g. @10am-3pm@
```

src/TzBot/Parser.hs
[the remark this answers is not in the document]
YuriRomanowski: I tried to avoid situations where time reference is parsed inside a piece of nonsense like "23djlk:5pm-asdqw8fl", and there is even one comment in another place, but you are right, it should be here too.

src/TzBot/Parser.hs
```diff
@@ -309,21 +372,37 @@ timeEntryParser = do
...
+              -> Matched (Maybe (TimeOfDay, Maybe $ Matched Bool), Bool -> TimeOfDay)
+              -> Matched TimeOfDay
+            applyIsAm isAm ref = do
+              let shouldAppend = isNothing $ getIsAm ref
+              whenFunc shouldAppend (modifyText (<> mtText isAm)) $ fmap (($ mtValue isAm) . snd) ref
+            extractDefaultResult :: Matched (Maybe (TimeOfDay, a), b) -> Maybe (Matched TimeOfDay)
+            extractDefaultResult ref = traverse (fmap fst . fst) ref
+        case isAmOptions of
+          [isAm] -> pure $ (TEPair `on` applyIsAm isAm) firstRef secondRef
+          _ -> maybe empty pure $ TEPair <$> extractDefaultResult firstRef <*> extractDefaultResult secondRef
```
YuriRomanowski, line 391: This code is very weird :laughing: Maybe I'll find how to make it more smart

YuriRomanowski:
In fact this PR ended up fixing all issues listed in #16 :smile:

dcastro:
> In fact this PR ended up fixing all issues listed in #16 smile

Sweet! :smile: I see you've also added `or` and `and` as possible delimiters! :D

dcastro:
Is this ready for review?

YuriRomanowski:
> Is this ready for review?

I think yes

dcastro (approved):

PR #82, opened 2023-02-24
Remind contexts and apply to time refs

YuriRomanowski:
## Description

Problem: Sometimes, context of a time reference can be spread over one sentence, several different sentences or even several different messages.

Solution: Parse contexts alone, remind them and apply to context-free time references when encountered; track a context during thread evolution.

## Related issue(s)

Part of #16 

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/src/TzBot/BotMain.hs
+++ b/src/TzBot/BotMain.hs
@@ -92,8 +92,12 @@ run opts = do
       managed $ withTzCacheDefault defaultMessageInfoCachingTime
     bsReportEntries <-
       managed $ withTzCacheDefault cCacheReportDialog
-    -- auto-acknowledge received messages
+
+    let defaultConversationStateCachingTime = hour 12
+    bsConversationStateCache <-
+      managed $ withTzCacheDefault defaultConversationStateCachingTime
     (bsLogNamespace, bsLogContext, bsLogEnv) <- managed $ withLogger cLogLevel
+    -- auto-acknowledge received messages
     liftIO $ runSocketMode sCfg $ handler gracefulShutdownContainer BotState {..}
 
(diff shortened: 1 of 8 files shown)


YuriRomanowski:
Seems much better :sweat_smile: 
изображение

src/TzBot/BotMain.hs
```diff
@@ -92,8 +92,12 @@ run opts = do
       managed $ withTzCacheDefault defaultMessageInfoCachingTime
     bsReportEntries <-
       managed $ withTzCacheDefault cCacheReportDialog
-    -- auto-acknowledge received messages
+
+    let defaultConversationStateCachingTime = hour 12
```
YuriRomanowski, line 96: This ideally should be in the config

src/TzBot/ProcessEvents/Message.hs
```diff
@@ -155,25 +153,52 @@ processMessageEvent' evt mEventType sender timeRefs =
...
+    (timeRefs, newState) <-
+      getTimeReferencesAndNewStateFromMessage conversationState msg
+    when (not $ null timeRefs) $
       -- save message only if time references are present
-      asks bsMessageCache >>= Cache.insert msgId timeRefs
+      asks bsMessageCache >>= Cache.insert msgId (timeRefs, newState)
+    Cache.insert getMessageThreadId (msgId, newState) convStateCache
+    asks bsMessageCache >>= Cache.insert msgId (timeRefs, conversationState)
+
+    withNonEmptyTimeRefs timeRefs $ \neTimeRefs -> do
```
YuriRomanowski, line 201: It seems that we will eventually rewrite this logic using some abstracting technique and write some tests, it gets a bit tricky as more internal state is kept.

PR #84, opened 2023-03-02
[Chore] Add deployment step to buildkite pipeline

Sereja313:
## Description

Problem: We need to set up CD.

Solution: Add deploy output to flake.nix, add deploy step to buildkite pipeline.
## Related issue(s)

https://issues.serokell.io/issue/OPS-1409

- Tests
        silently reappearing again.

- Documentation
    - Haddock

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -37,3 +37,10 @@ steps:
 
   - label: REUSE lint
     command: nix build -L .#checks.x86_64-linux.reuse-lint
+
+  # wait for all checks to finish
+  - wait
+
+  - label : deploy staging
+    command: nix develop .#ci -c deploy .#staging --ssh-user deploy --skip-checks
+    branches: staging
--- a/flake.nix
+++ b/flake.nix
@@ -23,9 +23,22 @@
     };
   };
 
-  outputs = { self, nixpkgs, haskell-nix, hackage, stackage, serokell-nix, flake-compat, flake-utils, ... }@inputs:
+  outputs = { self, nixpkgs, haskell-nix, hackage, stackage, serokell-nix, flake-compat, flake-utils, deploy-rs, ... }@inputs:
   {
     nixosModules.default = import ./module.nix inputs;
+
+    deploy.nodes = {
+      staging = {
+        hostname = "tejat-prior.gemini.serokell.team";
+        user = "deploy";
+        sshOpts = [ "-p 17788" ];
+        profiles = {
+          tzbot.path = deploy-rs.lib.x86_64-linux.activate.custom
+            self.packages.x86_64-linux.tzbot
+            "/run/current-system/sw/bin/systemctl restart tzbot";
+        };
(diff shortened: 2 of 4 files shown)


.buildkite/pipeline.yml
```diff
@@ -37,3 +37,10 @@ steps:
 
   - label: REUSE lint
     command: nix build -L .#checks.x86_64-linux.reuse-lint
+
+  # wait for all checks to finish
+  - wait
+
+  - label : deploy staging
+    command: nix develop .#ci -c deploy .#staging --ssh-user deploy --skip-checks --impure
```
rvem, line 45: Hmm, why is `--impure` needed here?

Sereja313: Oh, we really don't need it. Thank you!

flake.nix
```diff
@@ -23,9 +23,22 @@
     };
   };
 
-  outputs = { self, nixpkgs, haskell-nix, hackage, stackage, serokell-nix, flake-compat, flake-utils, ... }@inputs:
+  outputs = { self, nixpkgs, haskell-nix, hackage, stackage, serokell-nix, flake-compat, flake-utils, deploy-rs, ... }@inputs:
   {
     nixosModules.default = import ./module.nix inputs;
```
rvem, line 28: It might be useful to use [`ConditionPathExists`](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#ConditionPathExists=) for tzbot systemd service to ensure that `cfg.package` actually exists before starting the service. Somehow, I forgot about it when initially adding `nixosModule` for it

rvem: Without it, an attempt to deploy changes from https://github.com/serokell/gemini-infra/pull/76 will fail because `tzbot.service` will fail to start due to a lack of `/nix/var/nix/profiles/per-user/deploy/tzbot`

rvem: Note that it'll also require to update `tzbot` input in `gemini-infra` to use the newest version of `tzbot` module

.buildkite/pipeline.yml
```diff
@@ -37,3 +37,10 @@ steps:
 
   - label: REUSE lint
     command: nix build -L .#checks.x86_64-linux.reuse-lint
+
+  # wait for all checks to finish
+  - wait
+
+  - label : deploy staging
+    command: nix develop .#ci -c deploy .#staging --ssh-user deploy --skip-checks
```
rvem, line 45: I checked the encoded key in the Vault that is supposed to be fetched and imported to `ssh-agent` and it doesn't seem to have the right format. Could you please double-check it?

rvem (approved):
LGTM now:)

rvem:
@Sereja313, let's merge and proceed with the `gemini-infra` update :smile_cat:

.buildkite/pipeline.yml
rvem: Thanks, looks ok now

PR #129, opened 2024-03-05
[OPS-1161] Harden systemd service

Sereja313:
Problem: We want to harden the security of our systemd services.

Solution: Use the hardening profile defined in serokell.nix.

## Related issue(s)

Fixed #

- Tests
        silently reappearing again.

- Documentation
    - Haddock

## ✓ Release Checklist

- [ ] I updated the version number in `package.yaml`.
- [ ] (After merging) I created a new entry in the [releases](https://github.com/serokell/tzbot/releases) page,
      with a summary of all user-facing changes.
    *  I made sure a tag was created using the format `vX.Y`

### DIFF
--- a/module.nix
+++ b/module.nix
@@ -50,14 +50,32 @@ in
         export SLACK_TZ_BOT_TOKEN="${cfg.slackBotToken}"
         ${cfg.package}/bin/tzbot-exe --config ${pkgs.writeText "config.yml" (builtins.toJSON cfg.botConfig)}
       '';
+
       startLimitBurst = mkDefault 5;
       startLimitIntervalSec = mkDefault 300;
-      serviceConfig = {
+      serviceConfig = withHardeningProfile hardeningProfiles.backend {
         User = "tzbot";
         Group = "tzbot";
         StateDirectory = "tzbot";
         Restart = mkDefault "on-failure";
         RestartSec = mkDefault 10;
+
+        SystemCallFilter = [
+          "~@clock"
+          "~@debug"
+          "~@module"
+          "~@mount"
+          "~@raw-io"
+          "~@reboot"
+          "~@swap"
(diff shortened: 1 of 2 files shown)


Sereja313:
Depends on https://github.com/serokell/serokell.nix/pull/158

module.nix
```diff
@@ -50,14 +50,32 @@ in
         export SLACK_TZ_BOT_TOKEN="${cfg.slackBotToken}"
         ${cfg.package}/bin/tzbot-exe --config ${pkgs.writeText "config.yml" (builtins.toJSON cfg.botConfig)}
       '';
+
       startLimitBurst = mkDefault 5;
       startLimitIntervalSec = mkDefault 300;
-      serviceConfig = {
+      serviceConfig = withHardeningProfile hardeningProfiles.backend {
```
rvem, line 56: AFAIU, `hardeningProfiles.backend` implies `PrivateNetwork="yes"` which is too restrictive with tzbot since it needs to communicate with slack API servers

module.nix
```diff
@@ -50,14 +50,32 @@ in
...
         User = "tzbot";
         Group = "tzbot";
         StateDirectory = "tzbot";
         Restart = mkDefault "on-failure";
         RestartSec = mkDefault 10;
+
+        SystemCallFilter = [
```
rvem, line 63: Do you know if there is an easy way to determine which syscalls are needed by a given service?

rvem (approved):

module.nix
Sereja313: I dont think there is any easy way. You can use `strace` or  https://github.com/synacktiv/shh (which didn't work for me) but you will still need to manually use the service. Which is practically the same as simply restricting syscalls and then using the service to see if anything is broken

